PR #40, opened 2015-02-25
Added wrapper for Cli_GetPlcStatus

victorclaessen:
Added wrapper for Cli_GetPlcStatus 
and fixed get_block_info to allow input of block types other than 'data block'.


gijzelaerr:
do you know about `pull --rebase`? it will check in the changes and try to 'rebase' the tree, so you don't need to merge upstream changes into your branch.

victorclaessen:
Thanks for the hint, I suspected there would a better way of doing just that, but I wasn't able to find it.

PR #47, opened 2015-04-25
Update for S7-200/Logo Compatibility 

SimplyAutomationized:
Added snap7.client.Client().connect() for the Cli_Connect hook and unit tests addition

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -114,6 +114,13 @@ class Client(object):
             self.pointer, c_char_p(six.b(address)),
             c_int(rack), c_int(slot))
 
+    def connect(self):
+		"""To connect to these PLC you need to call set_connection_params() and then 
+		connect().  S7-200/Logo PLCs requiring TSAPs
+		"""
+		logger.debug("connecting using set_connection_params")
+		return self.library.Cli_Connect(self.pointer)
+
     def db_read(self, db_number, start, size):
         """This is a lean function of Cli_ReadArea() to read PLC DB.
 
@@ -523,7 +530,7 @@ class Client(object):
         type_ = param_types[number]
         return self.library.Cli_SetParam(self.pointer, number,
                                          byref(type_(value)))
-
+	
     def get_param(self, number):
         """Reads an internal Client object parameter.
         """
@@ -534,3 +541,6 @@ class Client(object):
                                          byref(value))
         check_error(code)
         return value.value
+
+
+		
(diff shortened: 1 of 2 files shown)


spreeker:
Awesome tanks for your commit. If it's possible run run the Test without actually having a logo 200 plc and run the tests against the virtual snap7 plc that would be awesome. Then the automated testing would work again.

SimplyAutomationized:
I ran the tests against the virtual snap 7 and connected successfully.  You also wanted me to test this with a PLC that isn't A S7-200/Logo?

spreeker:
Well Travis which is a service which runs all test is failing. 

    Error — The Travis CI build could not complete due to an error

The tests fail because all the tests using `connect` are now failing because you redefined the original connect function.  Python does not support method overloading. 
So you need to give your connect method a different name like `simple_connect` or `connect_logo`

What I meant by running the tests without a PLC is that in the tests should run standalone. The test should setup a virtual test environment and test against that. That is how all the other tests work.
Try to run all the test locally. Fix your own test to work in a similar matter and then we can commit your code.

below the offending code.
```

    @error_wrap
    def connect(self, address, rack, slot, tcpport=102):
        """
        Connect to a S7 server.
        :param address: IP address of server
        :param rack: rack on server
        :param slot: slot on server.
        """
        logger.info("connecting to %s:%s rack %s slot %s" % (address, tcpport,
                                                             rack, slot))

        self.set_param(snap7.snap7types.RemotePort, tcpport)
        return self.library.Cli_ConnectTo(
            self.pointer, c_char_p(six.b(address)),
            c_int(rack), c_int(slot))

    def connect(self):
		"""To connect to these PLC you need to call set_connection_params() and then 
		connect().  S7-200/Logo PLCs requiring TSAPs
		"""
		logger.debug("connecting using set_connection_params")
		return self.library.Cli_Connect(self.pointer)

```

SimplyAutomationized:
OK :-) will do.

gijzelaerr:
Hi All, 

We didn't implement `Cli_Connect` since it is the same as `Cli_ConnectTo` but with implicit parameters. You can also call `set_connection_params` to set parameters and then use `Cli_ConnectTo` (implemented as `Client.connect` in python) to connect to your PLC. 

So snap7 doesn't need any modification as far as I know. I'm closing this PR, please open an other issue if you have problems.

PR #49, opened 2015-06-16
implement as_b_send b_send and b_recv for Partner connection.

wilkpio:
Support for  b_send and b_recv in Partner mode.

### DIFF
--- a/snap7/partner.py
+++ b/snap7/partner.py
@@ -34,13 +34,21 @@ class Partner(object):
         self.pointer = None
         self.create(active)
 
-    def as_b_send(self):
+    def as_b_send(self, data):
         """
         Sends a data packet to the partner. This function is asynchronous, i.e.
         it terminates immediately, a completion method is needed to know when
         the transfer is complete.
         """
-        return self.library.Par_AsBSend(self.pointer)
+        rId = ctypes.c_int32(1)
+        size = len(data)
+        wordlen = snap7.snap7types.S7WLByte
+        type_ = snap7.snap7types.wordlen_to_ctypes[wordlen]
+        cdata = (type_ * size).from_buffer(data)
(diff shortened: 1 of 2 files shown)


gijzelaerr:
the travis test suite is failing, can you have a look?

https://travis-ci.org/gijzelaerr/python-snap7/jobs/67027448

wilkpio:
Does it execute tests for different platforms in parallel?
My experience with Partner mode is that there could be only one connection
at the time.
In case IP/Port/TSAP is already in use the connection will simply fail :(
Is it somehow possible to serialize tests for different platforms in travis?

br,
Piotr

best regards
Piotr Wilkosz
Disclaimer: Any errors in spelling, tact, or fact are transmission errors.

gijzelaerr:
Every tests suits is run in a VM, so they should not interfere.

gijzelaerr:
Sorry for the delay looking at this.

Your PR looks very good. But I'm trying to get this working on a mac, but I don't understand why it doesn't work. 

```
passive = snap7.partner.Partner(active=False)
active = snap7.partner.Partner(active=True)
passive.start_to("0.0.0.0", "127.0.0.1", 1002, 1000)
active.start_to("0.0.0.0", "127.0.0.1", 1000, 1002)
active.get_status().value
1 # (Running, active and trying to connect.)
passive.get_status().value
2 # (Running, passive and waiting for a connection)
```

gijzelaerr:
actually I'm seeing the same results on a Linux machine, they don't seem to connect and the test fails.

I'm trying this with snap7 1.4.0 on Ubuntu 15.04 and OSX 10.10.5.

gijzelaerr:
When this issue has been solved I'll release 0.5 and include these changes.

PR #55, opened 2015-11-12
Added Cli_GetPduLength function

MacBath:
Hey,

i added a wrapper for a client function to a fork of your python-snap7 project.
Guess it would be nice to see it in your repo. 

Hope thats the way to let you pull it. Never did something like this before.

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -534,3 +534,15 @@ class Client(object):
                                          byref(value))
         check_error(code)
         return value.value
+
+    def get_pdu_length(self):
+        """
+        Returns info about the PDU length.
+        """
+        logger.info("getting PDU length")
+        requested_ = c_uint16()
+        negotiated_ = c_uint16()
+        code = self.library.Cli_GetPduLength(self.pointer, byref(requested_), byref(negotiated_))
+        check_error(code)
+
+        return negotiated_.value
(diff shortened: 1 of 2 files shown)


snap7/client.py
```diff
@@ -534,3 +534,15 @@ def get_param(self, number):
                                          byref(value))
         check_error(code)
         return value.value
+
+    def get_pdu_length(self):
```
gijzelaerr: why not add requested and negotiated as arguments to the `get_pdu_length()` function?

gijzelaerr:
Hi @MacBath, thanks for your PR, you have done well. 

Please see my inline comment. Why don't you add requested and negotiated as arguments, does the function work as intended when you don't supply these?

MacBath:
Both Params will be set by the lib function Cli_GetPduLength. Even if i modify the requested PDU Size before, The function overrides the 1st Value with Param Nr. 10. And the 2nd Param with the negotiated PDU Size.
For my purposes it is enough to get the actual PDU size. if i need the requested, i can use the get_param function.

as to work as intended
Tested this code with a S7-1200 and with the Pure-C Example of snap7.

+-----------------------------------------------------
| UNIT Connection
+-----------------------------------------------------
| Result         : OK
| Execution time : 14 ms
+-----------------------------------------------------
  Connected to   : 172.24.66.100 (Rack=0, Slot=0)
  PDU Requested  : 480 bytes
  PDU Negotiated : 240 bytes

and the python function returns the same values.

gijzelaerr:
ok fine! merging this in. Thanks!

PR #57, opened 2016-01-14
Added Cli_GetCpuInfo function

jdugge:
Added wrapper and accompanying struct for the Cli_GetCpuInfo function.

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -89,6 +89,15 @@ class Client(object):
         logging.debug("CPU state is %s" % status_string)
         return status_string
 
+    def get_cpu_info(self):
+        """
+        Retrieves CPU info from client
+        """
+        info = snap7.snap7types.S7CpuInfo()
+        result = self.library.Cli_GetCpuInfo(self.pointer, byref(info))
+        check_error(result, context="client")
+        return info
+
     @error_wrap
     def disconnect(self):
         """
--- a/snap7/snap7types.py
+++ b/snap7/snap7types.py
@@ -229,3 +229,12 @@ class S7DataItem(ctypes.Structure):
         ('Amount', ctypes.c_int32),
         ('pData', ctypes.POINTER(ctypes.c_uint8))
     ]
+
+class S7CpuInfo(ctypes.Structure):
+    _fields_ = [
+        ('ModuleTypeName', ctypes.c_char*33),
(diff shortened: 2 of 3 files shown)


gijzelaerr:
hi @jdugge! Thanks for your contribution!

Two small issues:
 * Can you squash the 2 commits and force push the branch?
 * Can you add a test to the test suite to make sure the code works?

Thanks again!

jdugge:
Hi @gijzelaerr,

done!

Thanks for your work on maintaining this, it's really appreciated

gijzelaerr:
hi @jdugge awesome, thanks! There is a small problem with your code in combination with python3, you test between a byte string and a unicode string, typical Python3 incompatibility with Python2. Can you explicitly convert the returned to a string?

jdugge:
Done. I assume the conversion to unicode should be done in `test_get_cpu_info`, and not in the `get_cpu_info` function itself, because in real life, the user would want the original byte string.

Thanks for your patience! Let me know if anything else needs fixing. The next pull request will be smoother, I promise `=)`

spreeker:
Thanks @jdugge and @gijzelaerr for maintaining and make connection problems with siemens PLC something from the past! <3

gijzelaerr:
@jdugge no problems, you've done well.

About where to convert it, to me it looks like the returned values are always strings, why would you want to access it as bytes?

jdugge:
I was thinking one would want it returned as bytes because you have no way of knowing what encoding is used for storing the string in the PLC, so the decoding to unicode should happen on the user side. Granted, it's probably going to be straight ASCII most of the time, so in practice it probably won't make much difference either way, but if someone stores strings in something other than ASCII or UTF8, they might run into trouble.
It seems `get_block_info()` also and `get_cpu_state()` also return byte strings.

Maybe it would be best if functions that return strings always return unicode strings and take an optional `encoding` parameter that would default to `encoding='ascii'`? (Or to `encoding=locale.getpreferredencoding()` to be consistent with Python 3's `open()`, but that might make less sense in the context of PLCs)

gijzelaerr:
Ok, I think you are right! I'll just merge it. Thanks.

gijzelaerr:
i've made a 0.5 release which includes your PR! 

party time!

jdugge:
Yeah!

PR #69, opened 2017-02-22
Add __del__ to client, server and partner

ellepdesk:
I found this behavior by a mis-use of snap7, instead of just re-connecting on error I instantiated a new client as well. 

This eventually led to the last client not being able to open a socket anymore, as the maximum number of sockets for the user was reached.

This is of course better solved by not instantiating a new client, but I think it's good to let python-gc clean up the no-longer used instances as well.

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -31,10 +31,13 @@ class Client(object):
     A snap7 client
     """
     def __init__(self):
-        self.library = load_library()
         self.pointer = False
+        self.library = load_library()
         self.create()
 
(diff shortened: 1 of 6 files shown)


gijzelaerr:
awesome, looks good! Do you think there is a way to properly unittest it to make sure this actually does what it is supposed to do?

ellepdesk:
I added some unittests to test the *_Destroy functions are called on delete.

gijzelaerr:
perfect!

PR #72, opened 2017-03-01
Rewrite unittest to use assertEqual and assertIn

ellepdesk:
When tests fail using assertTrue does not provide information on the values
When using assertEqual and assertIn it will provide the expected and actual
values


gijzelaerr:
nice!

PR #73, opened 2017-03-01
Register PA area in snap7-server

ellepdesk:
ab_read() queries PA area on the server, ab_read() would fail because
data could not be read on the server

Adding PA area fixes test_ab_read

un-skip test_ab_read


gijzelaerr:
aha! great! :)

PR #75, opened 2017-03-02
corrected logger calls. 

tmcolby:
some of the modules were mistakenly calling the root logger (logging.debug or logging.info) instead of using the logger object that was declared in the module.


gijzelaerr:
yes! great! Thanks

PR #76, opened 2017-03-09
Fixed get_int bug

PeterTM:
get_int would fail with 'struct.error: pack expected 2 items for packing (got 0)' when a byte index other than 0 was supplied. 
Fixes:

- Test added to read an int from the end of the byte array
- data byte array end range changed to byte_index+2 instead of a fixed value of 2


gijzelaerr:
thanks!

PR #77, opened 2017-03-16
Permit the usage of write functions with read-only buffers

pauluap:
With Python3, when an byte string (e.g. b'\x00\x02') is iterated over, it provides a read-only view, and thus the usage of ctypes' from_buffer function in the write functions cause an 'underlying buffer is not writable' exception to be thrown.

e.g.

import snap7
c = snap7.Client()
c.connect('192.168.1.1', 0, 0)
c.write(464, b'\x00\x02')

This pull request modifies the write functions to use the from_buffer_copy function.  I suppose it's possible that there's a performance impact, but I expect it to be minimal given the capacity of the usual devices.

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -153,7 +153,7 @@ class Client(object):
         wordlen = snap7.snap7types.S7WLByte
         type_ = snap7.snap7types.wordlen_to_ctypes[wordlen]
         size = len(data)
-        cdata = (type_ * size).from_buffer(data)
+        cdata = (type_ * size).from_buffer_copy(data)
         logger.debug("db_write db_number:%s start:%s size:%s data:%s" %
                      (db_number, start, size, data))
         return self.library.Cli_DBWrite(self.pointer, db_number, start, size,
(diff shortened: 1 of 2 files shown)


gijzelaerr:
hi! Thanks @pauluap for this awesome pull request. Do you know why the test suite didn't bump into this error before? We test for Python3 and as far as I know it is covered by the unit test. If not, could you maybe add a unittest?

thanks!

pauluap:
I suspect that it's because the unit tests use bytearrays rather than byte string literals (can you tell I haven't looked at the unit tests yet?). I can try to add a few unit tests.  It seems that some of the code doesn't have unit tests in the first place.  I'll try to cover those for this particular case.

gijzelaerr:
@pauluap that would be awesome! Thanks already.

pauluap:
Allright, I added some unit tests.

My error, this isn't Python3-specific, Python2 also throws due to the use of string literals b'Something' is str() on Py2.7 and bytes() on Py3

I also elected to use mock to skip the actual call into the Snap7 C library.  This is a different methodology as compared to the rest of the client tests.  Would you have a problem with that?

gijzelaerr:
perfect! thanks a lot

gijzelaerr:
made a new release https://github.com/gijzelaerr/python-snap7/releases/tag/0.8

PR #82, opened 2017-08-01
Delete function supported

jon85p:
Hello
In a project I coded this tiny function (it worked a treat!), I want help this Python module with this little show of support.

PS = Thanks for this pretty nice module.


gijzelaerr:
hi @jon85p ! Thanks for your contribution. Can you maybe also add a unittest so we can automatically test if it works on all supported platforms? That would be perfect!

gijzelaerr:
let me just merge this then, lets see if it breaks!

PR #83, opened 2017-09-05
The method lock_area should call Srv_LockArea

oscarh:
The method lock_area currently calls Srv_UnlockArea.

I'm not very proud of the test, as it relies on waiting on thread and that one second is enough time to prove that a thread is blocked... Didn't come up with any better idea though. The test and the fix are in different commits. It should therefore be possible to merge just the fix if the test looks too horrible.

### DIFF
--- a/snap7/server.py
+++ b/snap7/server.py
@@ -212,7 +212,7 @@ class Server(object):
         """Locks a shared memory area.
         """
         logger.debug("locking area code %s index %s" % (code, index))
-        return self.library.Srv_UnlockArea(self.pointer, code, index)
+        return self.library.Srv_LockArea(self.pointer, code, index)
 
     @error_wrap
     def start_to(self, ip, tcpport=102):
--- a/test/test_server.py
+++ b/test/test_server.py
@@ -42,12 +42,26 @@ class TestServer(unittest.TestCase):
         self.assertRaises(Exception, self.server.get_mask, 3)
 
     def test_lock_area(self):
+        from threading import Thread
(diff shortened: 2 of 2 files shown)


gijzelaerr:
hi @oscarh, thanks for your contribution. That doesn't look to bad. Have you considered working with queues? You can put a message on a queue in one thread and then check the queue in an other thread.

oscarh:
I'm not sure I understand how a queue would help. What I want to prove is that the thread is blocked, as there is no API to check if an area is locked. I could use a queue or a condition etc. to communicate between the threads, but it's hard to test that the thread doesn't communicate...

gijzelaerr:
If a thread is blocked you can let a queue timeout right? so you let make a thread with a blocking operating, after that operation you put something in the queue. You run the thread, thread is blocked Do a get with a timeout on the queue from the other thread, that should timeout., then unblock the thread and see if there is something in the queue.

What do you think?

oscarh:
I don't see why it's a better solution than trying to join the thread with a timeout? The test thread won't be able to join the spawned thread until it has been unblocked, so that it can finish. The race condition still exists with a queue. With *the* race condition I'm referring to "How long is long enough to prove that the thread is blocked"? In the end, if the thread doesn't finish or the thread doesn't publish a message, is pretty much the same thing, right?

gijzelaerr:
hi @oscarh, sorry for the delay. I think you are probably right. 

I'll not delay this any further :) Thanks for you contribution!

PR #88, opened 2017-11-26
Create logo.py

openarchitect:
Initial version of Siemens Logo 7/8 support. Due to a very own handling of the logo family it make sense to create a own class, which only contains the working functions. In the next version more special function for the logo planned.


gijzelaerr:
hi @openarchitect! Thanks for your great contribution.

One thing, it looks like that a lot of code is actually shared with the client class. Isn't it an idea to inherit the logo class from the client class, and override the specifics? That way you get the overlapping implementation for free!

Issue #94, opened 2017-12-21
Writing to an INT memory, overwrites the memories below it


ildomarcarvalho:
Dear all,
When I try to write to an integer memory, it overwrites some memories below this int memory. Can you help me to find out if it is a bug or I just using the the snap7 incorrectly?

```python
import snap7

layout = """
      0       DB_VAR_1                    INT
      2       DB_VAR_2                    INT
      4.0     IN_Reserva_Camera           BOOL
      4.1     IN_Sinal_Iluminacao         BOOL
      4.2     IN_Robo_Trigger_Cam         BOOL
      4.3     IN_Robo_UR5_Inspecao        BOOL
      4.4     IN_Robo_UR5_Operacao        BOOL
      4.5     IN_Robo_UR5_Sinal_4         BOOL
      4.6     IN_Robo_UR5_Iniciar_ACK     BOOL
      4.7     IN_Reserva_2                BOOL
      5.0     IN_Sensor_Pressostato       BOOL
      5.1     IN_Reserva_Sensor_1         BOOL
      5.2     IN_Mon_BT_Emer_Painel_Pr    BOOL
      5.3     IN_Mon_BT_Emer_Reserva_1    BOOL
      5.4     IN_Mon_BT_Emer_Reserva_2    BOOL
      5.5     IN_Mon_Rele_BT_Emer         BOOL
      5.6     IN_Mon_Rele_Chave_Porta     BOOL
      5.7     IN_Reserva_3                BOOL
      6.0     IN_Sinal_MGB_Pos_Porta      BOOL
      6.1     IN_Sinal_MGB_Pos_Bolt       BOOL
      6.2     IN_Sinal_MGB_Pos_Trava      BOOL
      6.3     IN_Sinal_MGB_Diag_Saida     BOOL
      6.4     IN_Solic_Abert_Porta_MGB    BOOL
      6.5     IN_Sinal_Reset_MGB          BOOL
      6.6     IN_Reserva_4                BOOL
      6.7     IN_Reserva_5                BOOL
      7.0     IN_BT_Lig_Deslig_Comando    BOOL
      7.1     IN_BT_Pos_Inicial           BOOL
      7.2     IN_BT_Reset_Falha           BOOL
      7.3     IN_Cha_Manual_Automatico    BOOL
      7.4     IN_Reserva_6                BOOL
      7.5     IN_Reserva_7                BOOL
      7.6     IN_Reserva_8                BOOL
[...]

PR #96, opened 2018-01-04
set_int was just adding data to the array, moving everything else

ildomarcarvalho:
As I wrote in this issue (https://github.com/gijzelaerr/python-snap7/issues/94), when I tried to write to an INT memory, the memories below it were changed.

Turned out the problem was that the set_int was just adding the data to the array and moving everything else, instead of override the memory.

I tested here and works perfectly now.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -135,7 +135,7 @@ def set_int(_bytearray, byte_index, _int):
     # make sure were dealing with an int
     _int = int(_int)
     _bytes = struct.unpack('2B', struct.pack('>h', _int))
-    _bytearray[byte_index:2] = _bytes
+    _bytearray[byte_index:byte_index + 2] = _bytes
 
 
 def get_int(_bytearray, byte_index):


gijzelaerr:
ha! thanks! @spreeker what do you think?

ildomarcarvalho:
It seems like the errors aren't from the code, but from the permissions. I'm sorry if I'm wrong, but this is the first time I'm contributing to a open source project and all of this is kind of new to me :)

How can I help to fix this?

gijzelaerr:
hi @ildomarcarvalho , thanks for your contribution and sorry for the delay responding. I can't really verify this PR at the moment, but it would help if you could write a unittest that verifies this is an issue and that your PR solves this problem. Could you please do this? Let me know if you have questions.

spreeker:
This is correct. I noted the programming error. But since I do not have a
PLC anymore..

But this onliner now has the intended behaviour.

Stephan.

gijzelaerr:
ah well then i'll just merge it. thanks @ildomarcarvalho !

PR #159, opened 2020-05-18
dint type management

crusher83x:
get and set for dint type management,
test file


gijzelaerr:
looking good, thanks.

Issue #164, opened 2020-06-15
Add python asyncio wrapper around as_ methods


swamper123:
I got my station (S1200) last week and had some time to deal with some functions.
The sync functions work like expected, but the async functions receive garbage (mostly zeros).

While debugging I received sometimes the right values, so it seems to be an awaiting problem. I tried to just async these functions, but it seems to not change anything on it's behaviour. 

I will have a deeper look under the hood of the plain snap7 or try some workarounds, so that the wrapper may deal with it asyncly.

swamper123:
Okay, it seems that we (python-snap7) are not responsible for that.

If I understood the snap7 Code right the order is like this:
- create Request
- create Thread which starts the request
- wait for time x (?not sure? there seems to be Methods waitingFor and wait_forever)
- return anything or random(?not sure? otherwise we wouldn't receive an answer)

It looks like it is concurrent, but in another style as I thought. =/

From the Python side, I could imagine to turn these functions still into async and await the result of what ever happens inside there(because we are waiting and we want to be informed if something received). In such a scenario, we would maybe faster/more flexible.

If I got any response in the snap7 sourceforege forum, I will let you all know.

swamper123:
Could somebody may check async requests with other stations (I just have the 1200 one)? Maybe it is caused by the reduced support for this hardware.
Even better would be trying to change an as request into an asyncio function.
Something like this:

```
   async def as_db_read(self, db_number, start, size):
        """
        This is the asynchronous counterpart of Cli_DBRead.
        :returns: user buffer.
        """
        logger.debug("db_read, db_number:%s, start:%s, size:%s" %
                     (db_number, start, size))

        type_ = snap7.snap7types.wordlen_to_ctypes[snap7.snap7types.S7WLByte]
        data = (type_ * size)()
        result = await (self.library.Cli_AsDBRead(self.pointer, db_number, start,
                                            size,  byref(data)))
        check_error(result, context="client")
        return bytearray(data)
```

This would may help circleing the root of this evil. :^)

gijzelaerr:
I actually don't own a PLC, so I can't help you, unfortunately.

spreeker:
I don't think adding 'async' to snap7 library calls is useful.

A PLC is very simple does only one command each loop. And is many times
slower than a PC./ Python.

A PLC can only do one command at the moment. So if you do async you need to
queue up all the commands to the PLC. But then..

If you create a separate small non async service that talks to your PLC
[...]

Issue #157, opened 2020-05-08
Add type annotations


gijzelaerr:

swamper123:
Do you think about Support for Python 3.8 or older? This would be important, especially for future async functions ( I may will look for in the near future).

gijzelaerr:
yes

swamper123:
So Python 3.8?

gijzelaerr:
What is your question? The software supports python 3.8 and older. Did you try?

swamper123:
On which Version of Python3 do you want to build up on. 3.6? 3.7? 3.8? Another one?
This would be neccessary to know, especially for async functions, which base on asyncio (they changed a lot between 3.4 and 3.8).

gijzelaerr:
If we drop python2 support I would say keep support for Python 3.6+ since some people still run that and it not EOL yet:

https://devguide.python.org/#status-of-python-branches

swamper123:
I never have worked with Travis, so could you drop the Python2 tests there?

gijzelaerr:
you can configure travis using the `.travis.yml` file in the root.

swamper123:
With #169, #166 and #170 Support was dropped.

Stuff you may have to do @gijzelaerr :
- A new Version Tag should be added and CHANGE file should be updated.
- I don't know if Pypy is tracking github versions, so maybe an update there as well would be good.
Then we are good to go and ready for the future. 😎

gijzelaerr:
py2 support has been dropped, but no type annotations have been added, so I'm leaving this issue open. Uploads to pypi are automated with travis. 

I'm not fully convinced of a new release yet, since of the py2 support drop and no other major changes. I'm not fully convinced yet by the maturity of the async code also, and it only implements 2 functions. Maybe we should mark it experimental for now, and see if the community comes back with feedback. @spreeker what do you think of the async Implementation?

swamper123:
The async code will be extended in the future, no worries. Next PRs about that are in my pipeline. 😄 

The async code works now, because in the older versions hasn't been any receive check and even return wrong values (zeros) if the timeframe was to short! It is still stuff to do like implementing other async methods and a way to use the callback style like in `Cli_SetAsCallback()` (eg. via `add_done_callback` of Future objects.
[...]

PR #166, opened 2020-07-01
Make as_db_read asyncio able

swamper123:
Make as_db_read asyncio able for Python
Fix some PEP8 faults


swamper123:
I tested the PR with a SIEMENS S7-1200, other tester are welcome.

A fix client-test program would be:
```

IP = "" #  TheIPAddress
RACK = #  RackNumber
SLOT = #  SlotNumber

DB = #  DBNumber
START = #  StartNumber
SIZE = #  SizeNumber

async def main():
    client = snap7.client.Client()
    client.connect(IP, RACK, SLOT)
    try:
        while True:
            print(f"{await cient.as_db_read(DB, LEN, SIZE)"})
            # await asyncio.sleep(1)  #  for an output delay
    except:
        print("Something happened")
        client.disconnect()

asyncio.run(main())
```

swamper123:
For the record:
This was made with Python 3.8 and should be compatible down to Python3.6. This would fix #164 as well and can be used as blueprint for other async functions. Maybe it could be possible to use the callback-method with Python-Future objects, which would may boost the performance as well.

Travis test fails, because python2.7 hasn't got asyncio.

My offer is to create a sepperate function for higher Python Versions, which would benefit from that.
In general I would recommened to drop Python2.7 support, because the EOL is reached a long time ago. I will make a cleanup PR where Python2.7 is banned as good as possible, like in #157  mentioned.

snap7/client.py
```diff
@@ -500,7 +504,12 @@ def as_db_get(self, db_number):
         check_error(result, context="client")
         return bytearray(_buffer)
 
-    def as_db_read(self, db_number, start, size):
+    async def async_wait_loop(self):
```
gijzelaerr, line 507: I think we should move the python async functions to a separate module and keep access to the lower level functions also. So just leave the function as is, but put the python asyncio specific code into a separate submodule. What do you think?

swamper123: I can do that if you wish, but I don't know if the "old" wrapper would work properly, since nobody is checking if a result already returned and the actual code doesn't allow to check in one of the three ways that exist.

Even then, it would be still the same if I poll all the time or wait idle_time until a value is received. Only the Callback way would work, which would end in future objects (and more confusing I guess).

swamper123: I would add an optional parameter, which could be used to choose different ways of checking if a value received.
Standard would be `None ` like "use original code".

gijzelaerr: I think it is just much more clear if there is one module that is a one-to-one direct wrapper to the snap7 lib and then a separate submodule that does the fancy new asyncio stuff. Maybe people actually use those calls already and build their own logic around it, so there is also a backwards compatibility argument there.

swamper123: I have chosen to create another class in client.py for two reasons:
1) It is still a Client, so intuitive it could join client.py as well
2) The async Client should be a childclass from Client, to be (and stay) on the same level. The necessarry methods are overwritten.
3) I was afraid if I import `Client` and some dependencies may change in future, that these are not imported in the seperate module. I don't want to `import client *`. A linter may not even recognise that dependencies are missing.

swamper123:
@gijzelaerr Do you have any change wishes?

gijzelaerr:
also codecov is not happy since the coverage dropped, can you add a unit test for the method in your new class?>

swamper123:
I will commit a test in the next days (never had worked with codecov before, but I think I understand now what it does ^^).

snap7/client.py
```diff
@@ -624,3 +627,58 @@ def set_plc_datetime(self, dt):
         buffer[5] = dt.year - 1900
 
         return self.library.Cli_SetPlcDateTime(self.pointer, byref(buffer))
+
+
+class ClientAsync(Client):
```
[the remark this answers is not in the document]
swamper123: It's a matter of taste. I put it in the same file, so that other developer may work easier, because the `ClientAsync` is only overwriting a few methods at the moment. As well an async Client would be still a Client for me.

snap7/client.py
```diff
@@ -500,7 +504,7 @@ def as_db_get(self, db_number):
         check_error(result, context="client")
         return bytearray(_buffer)
 
-    def as_db_read(self, db_number, start, size):
+    def as_db_read(self, db_number, start, size, timeout):
```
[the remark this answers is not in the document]
swamper123: This shouldn't have been there. Thanks!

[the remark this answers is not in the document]
swamper123: Okay my linter said that the method arguments are different. So timeout is now optional with 1 second as standard.

swamper123:
Codecov seems to fail at Python2 lines, but because these will be erased later in other PR I will do soon, I would ignore it for the moment.

swamper123:
🎉  It worked 🎊 and passed 🎉  :)

test/test_client.py
```diff
@@ -121,7 +121,7 @@ def test_read_multi_vars(self):
         byte_to_value = [util.get_real, util.get_real, util.get_int]
 
         # unpack and test the result of each read
-        for i in range(0, len(data_items)):
+        for i, _ in enumerate(data_items):
```
gijzelaerr, line 124: sorry, but i don't think this code is an improvement

test/test_client.py
```diff
@@ -202,8 +202,8 @@ def test_get_cpu_state(self):
         self.client.get_cpu_state()
 
     def test_set_session_password(self):
-        password = 'abcdefgh'
-        self.client.set_session_password(password)
+        pw = 'abcdefgh'
```
gijzelaerr, line 205: also this is no improvement, i think if we would run pylint it would even say it is a bad variable.

test/test_client_async.py
```diff
@@ -0,0 +1,583 @@
...
+logging.basicConfig(level=logging.WARNING)
+
+ip = '127.0.0.1'
+tcpport = 1102
+db_number = 1
+rack = 1
+slot = 1
+
+
+class TestClient(unittest.TestCase):
```
gijzelaerr, line 26: it looks like you copied the whole test suite for the client module?

This is not required. You can only copy and modify the tests that are actually overriden in the class.

test/test_client.py
swamper123: It was claimed by Codefactor to use enumerate instead of range(len)) expression. So I changed it.

test/test_client.py
swamper123: This was claimed by Codefactor, because of a "hardcoded password".

test/test_client_async.py
swamper123: Fixed

gijzelaerr:
looking good, well done, thanks.

PR #168, opened 2020-07-02
[WIP]Upgrade to python3

swamper123:
Removing Python2 from the Project.


swamper123:
I am not sure why python2.7 has passed...it shouldn't have I guess.
As well I can't point on the problem here:

```
ERROR: test_ab_read (test_client.TestClient)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/gijzelaerr/python-snap7/test/test_client.py", line 43, in setUp
    self.client.connect(ip, rack, slot, tcpport)
 File "/home/travis/build/gijzelaerr/python-snap7/snap7/client.py", line 27, in f
    check_error(code, context="client")
  File "/home/travis/build/gijzelaerr/python-snap7/snap7/common.py", line 66, in check_error
    raise Snap7Exception(error)
snap7.snap7exceptions.Snap7Exception: b' TCP : Connection refused'
```
[This happens with a bunch of methods](https://travis-ci.org/github/gijzelaerr/python-snap7/jobs/704248917).
Is Travis not using sudo permissions? Or is my problem anywhere on my side?


It seems that the server hasn't even started:

```
ERROR: test suite for <class 'test_client.TestClient'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/suite.py", line 229, in run
    self.tearDown()
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/suite.py", line 352, in tearDown
    self.teardownContext(ancestor)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/suite.py", line 368, in teardownContext
    try_run(context, names)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/travis/build/gijzelaerr/python-snap7/test/test_client.py", line 39, in tearDownClass
    kill(cls.server_pid, 1)
ProcessLookupError: [Errno 3] No such process
```

And I even get the warning:
```
 "_ctypes/callbacks.c", line 315, in 'calling callback function'
TypeError: _remove() takes at most 2 arguments (3 given)
```

I would appreciate for any help with this.

swamper123:
Same trouble with master branch.

example/logo_7_8.py
```diff
@@ -1,26 +1,26 @@
...
-plc.connect("192.168.0.41",0x1000,0x2000)
+plc.connect(logo_ip, 0x1000, 0x2000)
 
 if plc.get_connected():
     logger.info("connected")
     
     # read I1 from logo
-    vm_address = ("V923.0" if Logo_7==True else "V1024.0")
-    print ("I1: " + str(plc.read(vm_address)))
+    vm_address = ("V923.0" if Logo_7 is True else "V1024.0")
```
gijzelaerr, line 22: this should just be == be in this case it probably doesn't matter much since it is a bool.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+mock
+python-snap7~=0.11
```
gijzelaerr, line 2: why add a requirements.txt file? and why add the library itself to it?

snap7/server.py
```diff
@@ -38,7 +38,9 @@ def __init__(self, log=True):
         self.create()
         if log:
             self._set_log_callback()
-    
+        self._callback = None
+        self._read_callback = None
```
gijzelaerr, line 42: Is this related to the python3 upgrade?

gijzelaerr:
About the failing tests, this is probably because the server segfaults. this is new and unexpected behaviour. Unfortunately, I don't have the time to investigate.

example/logo_7_8.py
swamper123: Because it is a bool value, this would be the better style. In the logical way it wouldn't matter.

snap7/server.py
swamper123: Both has been initiated outside of __init__ so I add them as None (for readability).
While I let a linter run my code, I fixed that on the fly. It would be a better Python style in general.

requirements.txt
swamper123: Ops, that shouldn't have happen....

swamper123:
Thx for the comments @gijzelaerr . Maybe I was a little bit to encouraged at once. I will tear stuff apart and PR step by step.

I will take a closer look at it and post my thoughts (without the swearwords 😄 ) loudly here. Perhaps some other may join me. :)

example/logo_7_8.py
gijzelaerr: as far is I know it works, this is not a better style. With `is` you check if the identity is equal, not the values. We are interested in the values here. It works with bools, but that doesn't mean it a better idea :)

gijzelaerr:
ok thanks for your PR, and sorry for being strict but I try to keep the project tidy. By now it is quite old so there are some old messy leftovers, so a linting and pep8 cleanup are appreciated also.

example/logo_7_8.py
swamper123: Alright, I will switch back '=='

swamper123:
Don't be sorry for that, it is my fault when I mess up with the work package 😄
As a junior dev, I still have to learn some stuff (and need more discipline with the PRs), but I will improve in future. 💪

PR #169, opened 2020-07-03
Reformat to f-strings where possible

swamper123:
Drop Python 2 string format style


snap7/client.py
```diff
@@ -344,8 +341,7 @@ def get_block_info(self, blocktype, db_number):
         if not blocktype:
             raise Snap7Exception("The blocktype parameter was invalid")
 
-        logger.debug("retrieving block info for block %s of type %s" %
-                      (db_number, blocktype))
+        logger.debug("retrieving block info for block {} of type {}".format(db_number, blocktype))
```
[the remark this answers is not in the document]
swamper123: It seems that `logger `doesn't support this style (even if I wish it would). Logger either use the old style with %s or .format(), so I went along with .format()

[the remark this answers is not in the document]
swamper123: I ran the tests and they said: 
grafik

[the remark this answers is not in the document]
swamper123: Or may that be just the Unittests doing blackmagic here?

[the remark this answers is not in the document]
swamper123: Well indeed there is some black magic happening, which I can't figure out at the moment. My other PC with a quick setup of logging works like a charm with f-strings.
Mysterious, but I will deal on monday with that. If anyone may know an answer, let us know. :)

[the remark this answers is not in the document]
gijzelaerr: f"" strings are converted to normal strings before being passed on to the function call, so it should be fine:
```In [1]: import logging                  

In [2]: logging.error(f"{logging}")     
ERROR:root:<module 'logging' from '/usr/lib/python3.8/logging/__init__.py'>
```

the error must be somewhere else.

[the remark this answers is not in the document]
swamper123: He just complains at the start/compile time with that, but later in the script he execute it without an error.
This is realy annoying and interessting at the same time. I may need a while to figure out what is happening here.

swamper123:
Reworked all loggers to f-string (and some concatinations of string), because it seems not to be the root of the evil.

snap7/client.py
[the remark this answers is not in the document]
swamper123: A these I have is that my _unittest_ may be a reason on my IDE. It seems to execute python2 instead of python3 and while executing in via terminal it shows other errors, than via IDE.
If there should be a problem, than travis would have complained about it (like in the OSX test) as well I guess. Well this will be my riddle, not the repo ones so feel free to merge. All typos are fixed. ;)

example/example.py
```diff
@@ -73,8 +73,8 @@ def show_row(x):
         data = get_db_row(1, 4 + x * row_size, row_size)
         row = snap7.util.DB_Row(data, rc_if_db_1_layout,
                                 layout_offset=4)
-        print 'name', row['RC_IF_NAME']
-        print row['RC_IF_NAME']
+        print(('name', row['RC_IF_NAME']))
```
gijzelaerr, line 76: why the double brackets? You can just give multiple arguments to the print functions, now it prints a tuple

snap7/bin/snap7-server.py
```diff
@@ -9,10 +9,11 @@
 import sys
 
 
-logging.basicConfig()
-logger = logging.getLogger()
+logging.basicConfig(level=logging.INFO)
```
gijzelaerr, line 12: now the logger is set to info 2x

snap7/server.py
```diff
@@ -7,6 +7,8 @@
 import snap7.snap7types
 from snap7.common import check_error, load_library, ipv4
 from snap7 import six
+import tracemalloc
+tracemalloc.start()
```
gijzelaerr, line 11: this should not be committed.

snap7/six.py
```diff
@@ -165,10 +165,10 @@ def __init__(self, six_module_name):
 
     def _add_module(self, mod, *fullnames):
         for fullname in fullnames:
-            self.known_modules[self.name + "." + fullname] = mod
+            self.known_modules[f"{self.name}.{fullname}"] = mod
```
gijzelaerr, line 168: if you drop python2  support you can drop the whole six.py module.

swamper123: I would do this in another PR, so that this PR is focused on the f-strings.

example/example.py
swamper123: Fixed

snap7/bin/snap7-server.py
swamper123: Deleted unnecessarry line

snap7/server.py
swamper123: Removed

snap7/six.py
gijzelaerr: but you already disable the py2 tests in this PR

swamper123: Is removed.

gijzelaerr:
perfect, almost there! Why latin-1 and not just ascii? just curious.

swamper123:
Well ... it was used in six.py in case it was Python3. So I just took it as it is.

gijzelaerr:
i'm not saying it is wrong, I'm just curious why there are different encodings used here and there.

swamper123:
To be honest, I have no clue why you chose latin-1 years ago, but I don't think it would make a problem using ascii here (if necessarry).

gijzelaerr:
I didn't choose that, this is the six module: https://pypi.org/project/six/

swamper123:
Oh ok. I just viewed git blame and haven't known that this is a project on it's own.
I will dig for a while what it does. What I found at the moment from the docs:

> Binary and text data
> 
> Python 3 enforces the distinction between byte strings and text strings far more rigorously than Python 2 does; binary data cannot be automatically coerced to or from text data. six provides several functions to assist in classifying string data in all Python versions.
> 
> six.b(data)
> 
>     A “fake” bytes literal. data should always be a normal string literal. In Python 2, b() returns an 8-bit string. In Python 3, data is encoded with the latin-1 encoding to bytes.
> 
>     Note
> 
>     Since all Python versions 2.6 and after support the b prefix, code without 2.5 support doesn’t need b().
>

I am not sure which encoding is necessarry for these ctype functions at the moment.

swamper123:
Hmm...it is still a riddle for me, but I will share what I know at the moment:

In Python2 have strings been the same as bytes and have been 8-bit large. This is not the case for Python3.
So six.b() returned the object for python2. In Python 3 this has to be encoded. Latin-1 has 8-bit characters and would be compatible to Python2 stuff.
So what is important for us? I don't know, because this stuff is passed for ctype stuff. It wouldn't be important for the address, but I don't know what encoding is used with PLCs ( just the first 128 ones or the whole latin1 plate?).

Fun fact, the latin-1 stuff is not sufficient:

[PEP393](https://www.python.org/dev/peps/pep-0393/) Says:

> The data and utf8 pointers point to the same memory if the string uses only ASCII characters (using only Latin-1 is not sufficient).

gijzelaerr:
for strings I dont think it matters that much, as long as you decode them the same way as you encoded them. As far as I know UTF-8 is a superset of LATIN-1 which is a superset of ASCII and they are all backwards compatible (you can parse an ASCII string with LATIN-1 and a LATIN-1 string with UTF-8). So if we encode and decode everything to and from UTF-8 it is probably fine.

swamper123:
Okay, I found a better explanation why latin-1 was used in [PEP3120](https://www.python.org/dev/peps/pep-3120/):

> In Python 2.0, the source encoding changed to Latin-1 as a side effect of introducing Unicode. For Unicode string literals, the characters were still copied literally from the source file, but widened on a character-by-character basis. As Unicode gives a fixed interpretation to code points, this algorithm effectively fixed a source encoding, at least for files containing non-ASCII characters in Unicode literals.
PEP 263 identified the problem that you can use only those Unicode characters in a Unicode literal which are also in Latin-1, and introduced a syntax for declaring the source encoding. If no source encoding was given, the default should be ASCII. For compatibility with Python 2.0 and 2.1, files were interpreted as Latin-1 for a transitional period. This transition ended with Python 2.5, which gives an error if non-ASCII characters are encountered and no source encoding is declared.

swamper123:
Jep, I will change that in a second.

swamper123:
Okay Codecov wants a logger lever on DEBUG, because every log below INFO was ignored and codecov would fail.
But hey, this should work as well now. :)

snap7/bin/snap7-server.py
```diff
@@ -8,10 +8,8 @@
 import snap7
 import sys
 
-
-logging.basicConfig()
-logger = logging.getLogger()
-logger.setLevel(logging.INFO)
+logging.basicConfig(level=logging.DEBUG)
```
gijzelaerr, line 11: i dont think we should change the default logging level for the server. I'm not sure why this would increase the code coverage number? These things are unrelated as far as I know.

swamper123: I thought that coverage wouldn't check the line if the logger level was set higher.

gijzelaerr:
Why does Codecov want debug level? Codecov is just a tool to ensure that we do proper unit test coverage. It should not dictate us to modify the code in unnecessary ways.

swamper123:
Okay, than this was a missunderstanding. Sorry.

snap7/bin/snap7-server.py
gijzelaerr: no worries, but this is untrue. Maybe some code inside the logging module will not be executed but that isn't included in our code coverage report.

gijzelaerr:
You can just ignore the codecov report in this case, there is a difference since in coverage but that is mostly just because the number of lines changed.

gijzelaerr:
merged, thanks.

swamper123:
All right, thanks for your help!

PR #170, opened 2020-07-17
Python3 format changes

swamper123:
Updates to be more PEP8 conform
Drop obsolete python2 and older python3 lines
Update introductions.rst
Small fix of self.pointer and missing self expressions in __init__
Update README

We should update the Version number as well, because this is a huge change.


snap7/logo.py
```diff
@@ -95,8 +98,8 @@ def read(self, vm_address):
         start = 0
         wordlen = 0
         logger.debug(f"read, vm_address:{vm_address}")
-        if re.match("V[0-9]{1,4}\.[0-7]{1}", vm_address):
-            ## bit value
+        if re.match("V[0-9]{1,4}\.[0-7]", vm_address):
```
gijzelaerr, line 101: why is this changed?

swamper123: {1} is a single repetition which is obsolete.
\. would make trouble as well, because it is an escape sequence for python strings. So such lines should have a prefix with r"".
It is a difference between r"\n" (backslash and n) and "\n" (newline).
But I wanted to do the regex stuff in another PR.

gijzelaerr: ok sure, just asking :) maybe a good idea to make sure it is unittested so we are sure the behaviour doesn't change.

swamper123: Will be made in the regex part. 😄

PR #171, opened 2020-07-17
Add asyncio able as_db_write method

swamper123:
Add unittest for as_db_write
Fix not supported as_check method, raise Warning now


swamper123:
Add fix for #157 like recommended from @spreeker

PR #172, opened 2020-07-21
Add more asyncio able methods with unittests

swamper123:
Add method to be more DRY like
Add more asyncio able methods incl. unittests

Only problem is, that I can't verify them on a hardware, but the pattern of how these methods work seems to be the same.

### DIFF
--- a/snap7/client_async.py
+++ b/snap7/client_async.py
@@ -1,7 +1,7 @@
 """
 Snap7 async client used for connection to a siemens7 server.
 """
-from ctypes import c_int, byref
+from ctypes import c_int, byref, c_byte
 
 import logging
 import asyncio
@@ -9,8 +9,20 @@ import asyncio
 import snap7
 from snap7.common import check_error
 from .client import Client
+from snap7.snap7types import buffer_type, buffer_size
+
 
 logger = logging.getLogger(__name__)
+
+
+def error_wrap(func):
+    """Parses a s7 error code returned the decorated function."""
+
+    def f(*args, **kw):
+        code = func(*args, **kw)
+        check_error(code, context="client")
+
+    return f
 
 
 class ClientAsync(Client):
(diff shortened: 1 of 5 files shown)


snap7/client_async.py
```diff
@@ -1,18 +1,30 @@
...
 import snap7
 from snap7.common import check_error
 from .client import Client
+from snap7.snap7types import buffer_type, buffer_size
+
 
 logger = logging.getLogger(__name__)
 
 
+def error_wrap(func):
```
gijzelaerr, line 18: why not import this from the other module?

snap7/client_async.py
```diff
@@ -24,15 +36,15 @@ def __init__(self):
...
+        None - pass, like sync method without a receive check
         1 - Poll if result is available, otherwise do something else
         2 - wait_idle, do something else and then wait until an answer receives (not implemented, better use 1)
         3 - Callback a method/function if the answer receives (not implemented)
         :param mode: Mode how an async answer shall be handled
         :return:
         """
-        if mode not in [None, 1, 2]:
-            logger.warning(f"{mode} is not a legit mode. Has to be None, 1 or 2!")
+        if mode not in [None, 1, 2]:
```
gijzelaerr, line 46: useless end of line spaces

snap7/client_async.py
swamper123: It is not imported in server.py, client.py and partner.py as well from anywhere.
Would util.py maybe a good place to store it there and then import it in these files?

snap7/client_async.py
swamper123: Will be fixed in next push

snap7/client_async.py
gijzelaerr: ah you are right, not sure why. Yeah, i think it could be a good one to centralize, but it is not crazy important.

swamper123: util.py seems to be for something else. I would leave it where it is at the moment and we should tidy it up in a later PR.

swamper123:
Before this PR should be merged, #174  should be merged, so that the tests are correct.

Issue #173, opened 2020-07-21
Tests issues, even if they pass.


swamper123:
Nearly all Tests for the Clients seems to fail, because the TCP Connection was refused.
This could be caused by the server, which is not running or is not reachable.

From Travis I got:

> INFO:snap7.server:creating server
> INFO:snap7.server:setting event callback
> INFO:snap7.server:registering area 5, index 1, size 100
> INFO:snap7.server:registering area 1, index 1, size 100
> INFO:snap7.server:registering area 4, index 1, size 100
> INFO:snap7.server:registering area 3, index 1, size 100
> INFO:snap7.server:setting server TCP port to 1102
> INFO:snap7.server:starting server on 0.0.0.0:1102
> Traceback (most recent call last):
>   File "_ctypes/callbacks.c", line 234, in 'calling callback function'
> TypeError: 'c_byte_Array_100' object is not callable
> INFO:__main__:2020-07-21 13:27:57 Server started

and an example of a Tests:

> 
> ERROR: test_ab_read (test_client.TestClient)
> ---------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/build/gijzelaerr/python-snap7/test/test_client.py", line 41, in setUp
>     self.client.connect(ip, rack, slot, tcpport)
>   File "/home/travis/build/gijzelaerr/python-snap7/snap7/client.py", line 25, in f
>     check_error(code, context="client")
>   File "/home/travis/build/gijzelaerr/python-snap7/snap7/common.py", line 66, in check_error
>     raise Snap7Exception(error)
> snap7.snap7exceptions.Snap7Exception: b' ISO : An error occurred during recv TCP : Connection timed out'
> -------------------- >> begin captured logging << --------------------
> snap7.client: INFO: creating snap7 client
> snap7.client: INFO: connecting to 127.0.0.1:1102 rack 1 slot 1
> snap7.client: DEBUG: setting param number 2 to 1102
> snap7.common: DEBUG: error text for 0xa006e
> snap7.common: ERROR: b' ISO : An error occurred during recv TCP : Connection timed out'
> --------------------- >> end captured logging << ---------------------

I see multiple IP adresses being used and I'm not sure if this is correct.

The server seems to run on 0.0.0.0 and port 102.
https://github.com/gijzelaerr/python-snap7/blob/f3ede30a119c9f3fd27501d0d01242152c81bd5e/test/test_server.py#L114

[...]

PR #174, opened 2020-07-21
repair tests

swamper123:
Set logger level to WARNING
Change IP adresses for server, partner and clients
Mentioned in #173


swamper123:
Fix for #173

swamper123:
I had a deeper look in older tests and it seems that the osx tests throw errors as well and show similar behaviour of failing.
Because I neither got a mac nor work with that, I can't/won't make a PR for that.

swamper123:
Well..if the test is not included in the scruot it can't be tested. And of course some stuff happened so back to fixing again.

swamper123:
So this seems to be fine.

.travis/before_install.sh
```diff
@@ -4,8 +4,8 @@ set -x
 
 if [ "$TRAVIS_OS_NAME" = "osx" ]; then
     brew install snap7
-    pip install nose mock coverage
-    pip3 install nose mock coverage
+    pip install nose mock coverage aiounittest
```
gijzelaerr, line 7: i think this can be removed now since we dropped py2 support. the pip install in the script should work afaik.

run_tests.sh
```diff
@@ -1,12 +1,14 @@
 #!/bin/sh
 
 export PYTHONPATH=.
-echo "Running test_partner.py ----------------------------------------------\n\n"
+printf "Running test_partner.py ----------------------------------------------\n\n"
```
gijzelaerr, line 4: why is printf better than echo?

snap7/server.py
```diff
@@ -37,8 +37,6 @@ def __init__(self, log=True):
         self.create()
         if log:
             self._set_log_callback()
-        self._read_callback = None
-        self._callback = None
```
gijzelaerr, line 41: why remove these? they are set later on right?

run_tests.sh
swamper123: echo can behave different on different systems. There can be different interpretations, especially with `\n`, while printf can handle that.

snap7/server.py
swamper123: I had the problem, that if I add them into `__init__` (like it should in Python), the server would crash. This was the root of all evil, why all tests failed so hard. I don't know why and I don't know where this behaviour is caused.

.travis/before_install.sh
swamper123: Will be removed on next commit.

snap7/server.py
swamper123: Could it be caused by the wrapper function? I haven't dealed with function definitions in functions.

swamper123:
Like mentioned in #176 this PR can either be merged, so that the linux tests run again (I can rename it if @spreeker or @gijzelaerr would accept that) or we have to wait until somebody with Travis and osx knowledge help us out to fix a longer existing bug.

gijzelaerr:
Another solution would be to revert your changes.

snap7/server.py
swamper123: Without `self._read_callback` and `self._callback`, it would be the same like before my changes. They haven't exist before, but seem to let it running ... somehow.

What I can't understand is, that these attributes exist 3 times each in the project:
1) in '__init__'
2) in `set_read_events_callback` or `set_read_events_callback` where these are declared
3) as parameter in the last line of these upper methods.

I can't see a problem here from the python side. If something is self, it should be in '__init__', even if it is just None (like a good Python style). Python would even be so flexible to add these in the object, if it is declared in a method.

I can only imagine, that it can be a size problem and pointer magic. If I declare them as None, it gets an address where this variable is stored. If some c-type pointer magic may happen, which is looking for the address, it will find this space, but maybe the space size is not right. And this may happens before the interpreter (if he even notice) can handle that. 

This is the only thing I can imagine.

gijzelaerr:
you were adding the 


```
        self._read_callback = None
        self._callback = None
```

In the wrong location, you were overriding th callback function which was set by `_set_log_callback()` in server.py.

this PR should fix a lot of issues:

https://github.com/gijzelaerr/python-snap7/pull/182

PR #178, opened 2020-07-27
[WIP] Add type annotations for client and client_async

swamper123:
Start adding type annotations like mentioned in #157 .

Helping hands are welcomed :)

### DIFF
--- a/snap7/__init__.py
+++ b/snap7/__init__.py
@@ -4,6 +4,7 @@ The Snap7 Python library.
 import pkg_resources
 
 import snap7.client as client
+import snap7.client_async as client_async
 import snap7.common as common
 import snap7.error as error
 import snap7.logo as logo
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -6,6 +6,7 @@ import re
 from ctypes import c_int, c_char_p, byref, sizeof, c_uint16, c_int32, c_byte
 from ctypes import c_void_p
 from datetime import datetime
+from typing import Union, Optional
 
 import snap7
 from snap7.common import check_error, load_library, ipv4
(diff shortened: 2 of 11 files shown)


gijzelaerr:
cool, thanks. Maybe we should add some docker mypy run to travis also, similar to this project:

https://github.com/eduvpn/python-eduvpn-client/blob/master/docker/eduvpn_ubuntu_bionic_py3_mypy.docker

and when we are at it, also a pycodestyle:

https://github.com/eduvpn/python-eduvpn-client/blob/master/docker/eduvpn_ubuntu_bionic_py3_pep8.docker

that way we are for sure no broken code gets checked into the repo (anymore).

swamper123:
@gijzelaerr I never worked with dockers before (and not the time at the moment), so I will not be a great help in that case.

But what would mypy run do, compared to the other existing stuff in travis? What would be the advantage?

Pycodestyle can be usefull, but wouldn't pylint be better to check styles?

gijzelaerr:
i'm also quite busy so i have very little time for this kind of things, but answering your questions between work is possible for me.

having a docker container on Travis enables us to run these tools for every PR, so we/others don't need to do it manually. 

pylint is much more pedantic than pycodestyle, which is not necessarily a bad thing but it will take more time to adjust the code to its likings. pycodestyle is a good start to check for pep8 compatibility.

swamper123:
@gijzelaerr Hmm...if I make a PR, Travis starts on it's own running all its stuff. I never had to push a button or similar, no clue if other contributors have such a problem.

Well we could start with pylint as well. Something is better than nothing. 😄

gijzelaerr:
That travis runs on every PR is the whole point right? That way we check of the PR is good. so if we do a mypy and pycodestyle for every PR we are sure no bad code ends up in master.

swamper123:
In Case somebody wants to help, I want to mentioned some hints/cases:
- f-strings and bytes may bite. Perhaps `fr` prefix may fix it [Conclusion: TBH]
- Check if methods are equal to parent class
- --> Union[type1, type2, ...] says explicit that type1, type2 ... can be returned
- --> Optional[type1] says, that type1 or None can be returned

PR #180, opened 2020-07-28
Add as area read write

swamper123:
Add async read_area and async write_area to client.py and client_async.py

This will be a draft until #174 is merged.


snap7/client_async.py
```diff
@@ -173,3 +173,59 @@ async def as_download(self, data, block_num=-1, timeout=1):
...
+        :param size: number of units to read
+        """
+        assert area in snap7.snap7types.areas.values()
+        if area == snap7.snap7types.S7AreaTM:
+            wordlen = snap7.snap7types.S7WLTimer
+        elif area == snap7.snap7types.S7AreaCT:
+            wordlen = snap7.snap7types.S7WLCounter
+        else:
+            wordlen = snap7.snap7types.S7WLByte
+        type_ = snap7.snap7types.wordlen_to_ctypes[wordlen]
```
gijzelaerr, line 194: i think you dont need to replicate the code from client.py but just call it and make a python async wrapper around it.

snap7/client.py
```diff
@@ -616,6 +616,54 @@ def set_plc_datetime(self, dt):
 
         return self._library.Cli_SetPlcDateTime(self._pointer, byref(buffer))
 
+    def as_read_area(self, area, dbnumber, start, size):
```
gijzelaerr, line 619: i think i added a as_read_area() also in my cleanup (with a raise NotImplemented exception)

test/test_client_async.py
```diff
@@ -73,3 +73,52 @@ async def test_as_db_get(self):
     async def test_as_download(self):
         data = bytearray(128)
         await self.client.as_download(block_num=-1, data=data)
+
+    async def test_as_read_area(self):
+        amount = 1
+        start = 1
+        # Test read_area with a DB
+        area = snap7.snap7types.areas.DB
```
gijzelaerr, line 81: now the non-async one is not tested?

snap7/client_async.py
swamper123: I'm not sure if a wrapper is the solution, because check_error() would run twice with perhaps two different solutions.
In the sync version, the result is never awaited (just hoped that a legit result received in the time between the lines), so I'm not sure what may could happened there.
But you are still right, it's still not fitting in the DRY (don't repeat yourself) -concept, I will think about a better solution about that.

gijzelaerr: why would check_error run twice?

swamper123: Afaik if you wrap a function, the wrapper can perform prolog and epilog stuff only. So if I would wrap `as_area_read()` from `client.py` I would dive through the wrapper, to the function, execute it all way, do the epilog stuff (which could be the asyncio stuff) and exit it the whole statementblock again.
But `as_area_read()` already contain `check_error(result, context="client")` and adding       
```
 request_in_time = await self.as_check_and_wait(timeout)
 if request_in_time is False:
     return None
 return bytearray(data)
```
in the epilog would be `check_error()` again.

Or do you tend to something else?

snap7/client.py
swamper123: There is as_read_area and asreadarea. So you want them with a NotImplemented exception?

snap7/client_async.py
gijzelaerr: in the end, you just do a call to the library that you want to have error-checked right? You can use the existing library structure for that. Your async code is a wrapper around that.

snap7/client.py
gijzelaerr: oops, o forgot to make those methods `as_read_area` format, my bad. No, I made all those NotImplemented methods as a reminder of which methods still need to be implemented, you can remove that one if you add it now.

swamper123: As well, how would you like the names to be dealed? The already implemented methods have underscores (as_ab_read) and some not (asebread). I prefere the underscore version for these kind of methods, because it is easier to read. Other methods like the `async_wait_loop()` could be without underscores and are still readable.

snap7/client_async.py
swamper123: I see still a difference (and not good wrappable solution) between the sync:
```
        result = self._library.Cli_AsReadArea(self._pointer, area, dbnumber, start, size, wordlen, byref(data))
        return bytearray(data)
```
and my async version:
```
       result = self._library.Cli_AsReadArea(self._pointer, area, dbnumber, start, size, wordlen, byref(data))
        request_in_time = await self.as_check_and_wait(timeout)
        if request_in_time is False:
            return None
        return bytearray(data)
```
In the first example you are sending the read command, and then execute check_error directly afterwards. No check if AsReadArea was received.

In my example you are sending the request, wait until a timeout/execution and then check_error.

This leads to two different `results` (one checked/legit result, one not checked/faulty/perhaps right result)  and so `check_error()` may return different results, wouldn't it? 
I want explicitly wait until `CheckAsCompletion()` shows green light for a valid result, before I want to execute check_error(), because your result and my result will mostly be different.

If I would just make a decorator at `as_area_read()`,  the first `check_error()` may throw a Snap7Exception, even if I wait until a legit result afterwards in my wrapper code and the second `check_error()` there would be fine with that.

I will commit an idea the next hours, perhaps we will agree on that. :smile:  
Sorry for all the trouble I cause here.

snap7/client.py
```diff
@@ -625,6 +625,13 @@ def as_read_area(self, area, dbnumber, start, size):
         :param start: offset to start writing
         :param size: number of units to read
         """
+        wordlen, type_, data = self._as_read_area_prepare(area, dbnumber, start, size)
+        result = self._library.Cli_AsReadArea(self._pointer, area, dbnumber, start, size, wordlen, byref(data))
+        check_error(result, context="client")
+        return bytearray(data)
+
+    @staticmethod
+    def _as_read_area_prepare(area, dbnumber, start, size):
```
swamper123, line 634: This is the beginning of the same code in both cases, just as private method and returns the stuff which is needed.
Same in the as_write_area case.

snap7/client.py
gijzelaerr: underscore version is preferred.

snap7/client_async.py
gijzelaerr: you should always do the check_error on the result of the call, since the library will return an error code anyway. This will not change. Do not await in between.

swamper123: Sorry for that stupid question: So just the call itself is checked with `check_error()`, not the received data from the PLC? Because if this is the case, I would finally understand your view and then, of course, a wrapper could be the most elegant solution.

gijzelaerr: yes, the check_error() parses an error code returned by the call. You should always check if the low level call is successful.

swamper123:
*sigh* I just found another problem in my Code if I would do it like this:

While all the async write stuff will work, because it just writes and check if write was succesfull, all read/get methods will cause trouble.
If I do the stuff like in [093d80d], it may looks nice, but it causes SIGSEGV faults in praxis.

I suggest this is caused in the client class:
- I return `bytearray(data)` from the `client.py method.
- Then I wait in the `client_async.py` method, until a value is received for data.
- -->But the address of data became invalid after escaping the `client.py` method.
This causes either: A wrong value returns or SIGSEGV

snap7/client_async.py
```diff
@@ -3,27 +3,34 @@
...
+    else:
+        def f(*args, **kw):
+            code = func(*args, **kw)
+            check_error(code, context="client")
 
     return f
 
 
-class ClientAsync(Client):
+class ClientAsync(Client, ABC):
```
gijzelaerr, line 33: this is not how Abstract Base Classes work.

swamper123:
So async unittests are not supported for python3.6 and figuering out how this stuff works with pytest was a bit tricky. In an later PR I would try to optimise the code a better way, but this should be a more solid base at the moment.

swamper123:
This also fixes #183

gijzelaerr:
I was thinking a bit about the async code inheriting and overriding methods for a while, and I'm not fully sure if it is actually a good idea if the return type changes since it violates the Liskov Substitution Principe:

https://en.m.wikipedia.org/wiki/Liskov_substitution_principle

Now I'm not really a fan of OO anyway, but maybe it is a good idea to stay SOLID if we use it

https://en.m.wikipedia.org/wiki/SOLID

swamper123:
My goal was to be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), but if you like, I can seperate them again.

gijzelaerr:
I don't think you are not dry if it is a proxy for the supported methods, although I'm not sure what good OO design principles are for proxies that change the type signature.

PR #182, opened 2020-07-28
A big repo cleanup

gijzelaerr:
* Enable docker builds for various targets, including mypy, pycodestyle and sphinx doc build
* Remove old tox and nosetest stuff
* Add helper Makefile
* Remove old vagrant files
* Formatting fixing
* Depricate mock module, use unitest.mock


snap7/__init__.py
```diff
@@ -3,14 +3,13 @@
 """
 import pkg_resources
 
-import snap7.server as server
 import snap7.client as client
```
swamper123, line 6: I think there is the async client missing here.

snap7/server.py
```diff
@@ -287,3 +289,25 @@ def clear_events(self):
         """
         logger.debug("clearing event queue")
         return self.library.Srv_ClearEvents(self.pointer)
+
+
+def mainloop(tcpport: int = 1102):
```
swamper123, line 294: May be a stupid question, but why is the mainloop running with port 1102? IsnÄt 102 the S7comm used port?

snap7/__init__.py
gijzelaerr: it should first be a bit more mature

snap7/server.py
gijzelaerr: because then you dont need to run the test suite as root.

swamper123:
Did the Tests even start running? I can't see a result like [here](https://travis-ci.org/github/gijzelaerr/python-snap7/jobs/712819106) and other cases.

gijzelaerr:
there are 3 travis builds that have a NOSCRIPT environment variable, the script.sh bash script checks if that is set and then doesn't execute `make test`

gijzelaerr:
so the actions are actually performed in the install script, where a docker build is triggered. you can also do that yourself using makefile

PR #200, opened 2020-10-27
Add as area read write

swamper123:
Add comperation of values to test_write_area and test_read_area
Add as_write_area and as_read_area with tests for client.py


snap7/client_async.py
```diff
@@ -147,9 +138,10 @@ async def as_db_get(self, db_number, timeout=1):
         """
         logger.debug(f"db_get db_number: {db_number}")
         _buffer = buffer_type()
-        result = self._library.Cli_AsDBGet(self._pointer, db_number, byref(_buffer), byref(c_int(buffer_size)))
-        request_in_time = await self.as_check_and_wait(timeout)
-        if request_in_time is False:
+        result = self._library.Cli_AsDBGet(self._pointer, db_number,
```
gijzelaerr, line 141: This call should be in `snap7.client`

test/test_client_async.py
```diff
@@ -9,67 +9,124 @@
...
-        await self.client.as_ab_write(start=start, data=data)
-
-    async def test_as_db_get(self):
-        await self.client.db_get(db_number=db_number)
-
-    @unittest.skip("TODO: RuntimeWarning: coroutine 'TestClient.test_as_download' was never awaited")
-    async def test_as_download(self):
-        data = bytearray(128)
-        await self.client.as_download(block_num=-1, data=data)
+@pytest.fixture(scope='module')
```
gijzelaerr, line 22: i don't really like the pytest fixture syntax, I've run into issues with it in other projects, I think there is too much magic happening. What is the advantage above a normal test class?

test/test_logo_client.py
```diff
@@ -26,7 +28,7 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
-        kill(cls.server_pid, 1)
+        kill(cls.process, 1)
```
gijzelaerr, line 31: nice

snap7/client_async.py
```diff
@@ -77,29 +81,17 @@ async def as_db_write(self, db_number, start, data, timeout=1):
...
         """
-        wordlen = snap7.types.S7WLByte
-        type_ = snap7.types.wordlen_to_ctypes[wordlen]
-        size = len(data)
-        cdata = (type_ * size).from_buffer_copy(data)
-        logger.debug(f"db_write db_number:{db_number} start:{start} size:{size} data:{data}")
-        check = self._library.Cli_AsDBWrite(self._pointer, db_number, start, size, byref(cdata))
-        request_in_time = await self.as_check_and_wait(timeout)
-        if request_in_time is False:
+        check = Client.as_db_write(self, db_number, start, data)
```
gijzelaerr, line 84: see this already starts to look much cleaner, and if we want to switch to cffi at some point we only need to modify the client class.

snap7/client_async.py
```diff
@@ -3,22 +3,28 @@
...
 logger = logging.getLogger(__name__)
 
 
 def error_wrap(func):
     """Parses a s7 error code returned the decorated function."""
 
-    def f(*args, **kw):
-        code = func(*args, **kw)
-        check_error(code, context="client")
+    if asyncio.iscoroutinefunction(func):
```
gijzelaerr, line 19: i think i would make an explicit async error wrap function and use that for async functions, since you already know it is going to be a async function while programming

snap7/client.py
```diff
@@ -643,6 +643,60 @@ def wait_as_completion(self, timeout: int) -> int:
...
+        :param size: number of units to read
+        """
+
+        wordlen, data = self._as_read_area_prepare(area, size)
+        logger.debug(f"reading area: {area} dbnumber: {dbnumber} start: {start}: amount {size}: wordlen: {wordlen}")
+        result = self._library.Cli_AsReadArea(self._pointer, area, dbnumber, start, size, wordlen, byref(data))
+        check_error(result, context="client")
+        return data
+
+    @staticmethod
```
gijzelaerr, line 662: this class is already quite huge, maybe just move this out of the class or into utils

setup.py
```diff
@@ -4,7 +4,7 @@
 
 __version__ = "0.11"
 
-tests_require = ['pytest', 'mypy', 'pycodestyle']
+tests_require = ['pytest','pytest-asyncio', 'mypy', 'pycodestyle']
```
gijzelaerr, line 7: what does the pytest async module do that we cant do ourselves?

gijzelaerr (changes requested):

swamper123:
Oh dam, it seems, that I mixed up some commits here. Sorry for that.

The `client_async` and `test_client_async` stuff is deprecated, until the as_* requests are working okay in `client.py ` and are accepted in master. Because then this all gets much more prettier. I'm on it.

test/test_client_async.py
swamper123: `Unittest ` doesn't support asyncio tests in Python <=3.7. So the next best thing I found was pytest. It is hard to understand at first, but when you are into it, it is understandable what is happening with all the fixture stuff etc. I haven't had much problems, since I use it (but well, I guess you made more projects than I ever did). What kind of issues happened in your projects?

snap7/client.py
swamper123: I guess utils will be a good place. Will be done.

setup.py
swamper123: pytest-asyncio is a good plugin for pytest to handle Coroutines. I haven't found such a thing for Unittest (and Unittest is supporting asyncio stuff only >=Python3.8).

snap7/client.py
gijzelaerr: btw, if this is the only place where it is used then why not just put it in the function, no reason to split things up too much.

test/test_client_async.py
```diff
@@ -9,67 +11,55 @@
...
-        data = bytearray(size)
-        check = await self.client.as_db_write(db_number=1, start=0, data=data)
-        self.assertEqual(check, 0)
+@pytest.fixture(scope='module')
+def testserver():
+    process = Process(target=mainloop)
+    process.start()
+    time.sleep(2)  # wait for server to start
+    yield process
+    kill(process.pid, 1)
```
gijzelaerr, line 30: doesn't this directly kill the server again

test/test_client_async.py
```diff
@@ -9,67 +11,55 @@
...
+    client_async.destroy()
 
-    async def test_as_db_get(self):
-        await self.client.db_get(db_number=db_number)
 
-    @unittest.skip("TODO: RuntimeWarning: coroutine 'TestClient.test_as_download' was never awaited")
-    async def test_as_download(self):
-        data = bytearray(128)
-        await self.client.as_download(block_num=-1, data=data)
+async def test_wait_loop(testclient):
```
gijzelaerr, line 42: so what I dont like about this weird pytest fixture is that this `testclient` arguments comes out of nowhere, you have no way to quickly see where it comes from. I guess pytest keeps an internal storage of named fixtures you can use? How can you quickly jump to the function definition?

test/test_client_async.py
swamper123: This kills the server, if the test has finished...but I should have used the functional scope here.

test/test_client_async.py
swamper123: I understand what you mean. It isn't marked as pytest_testclient or similiar to see directly what and where it might come from (well you can name them as these) and depending on the IDE (if even used), this may be confusing. I agree with that.
 
There are two ways (I know) how to organise pytest fixtures (stuff you want to reuse in multiple places):
1) They are already existing in the test_file_XYZ itself, marked with a `@pytest.fixture(scope="funciton/module/class/session")`, depending on your testing purposes.

2) conftest.py - pytest is looking for this file in the directory structure. It is a convention in pytest. Each conftest.py provides configs for the file tree pytest finds it in. You can store the wanted fixtures (and other pytest related stuff) in it. So in our case one conftest.py file in the root directory would be enough. 

3) You can mark individual tests with `@pytest.mark.usefixtures('testclient')` (I learned today). So the confusion would be reduced about that. 
Or you can mark all tests in a file with `pytestmark = pytest.mark.webtest`.
If you need multiple marks for all tests: `pytestmark = [pytest.mark.webtest, pytest.mark.asyncio]`

So there are places existing to look for, if you know that pytest is used. It is handy to avoid boilerplate, especially when you want to reuse/parametrise tests.

test/test_client_async.py
gijzelaerr: so this is only yielded at the end of the test?

swamper123: It's like the setUp/tearDown way in unittest.
"setUp" is everything before the yield and "tearDown" everything behind it.

test/test_client_async.py
```diff
@@ -1,6 +1,8 @@
+import asyncio
+import ctypes
 import logging
 import time
-import unittest
+import pytest
```
swamper123, line 5: I guess this is causing mypy to crash. I don't know how this new test stuff works, but it looks like an error with the scope (pytest can't be found, because it is in the pytest file in venv??)

gijzelaerr:
about the mypy test fail, you need to add a module ignore statement to setup.cfg in the mypy section.

about the windows fail, my guess is it has something to do with running the server in the background and not stopping/killing/closing that properly, and somewhere internally the C lib does something illegal because of that. 

It's a bit painful since it is in a multiprocess environment, which makes it hard to figure out. But, this is unrelated to your PR, so I'm willing to just accept this PR once the mypy test is fixed. We can also just ignore the windows test for now. by adding a `continue-on-error` flag, like https://github.com/MonetDBSolutions/MonetDBe-Python/blob/master/.github/workflows/windows.yml#L11

swamper123:
> about the windows fail, my guess is it has something to do with running the server in the background and not stopping/killing/closing that properly, and somewhere internally the C lib does something illegal because of that.

This may be it. I will try to add that in my schedule this week and try some experiments in the "experimental draft" about that. Perhaps I can archive something so that everything gets right again.

gijzelaerr (approved):

PR #204, opened 2020-10-29
Switch decorator from unittest to pytest, since tests are run by pytest

swamper123:


swamper123:
Okay, I tried out some stuff, but I guess there is a Windows breaking thing, since the code seems to be okay and no kind of Python error was thrown in test_partner. =/

PR #210, opened 2020-11-12
Try catching SEGFAULTS and SIGABRT signals correctly

swamper123:
Changing the async methods to be more snap7 like
Adding _prepare functions to create right wordlength and usrdata reference
Rework some tests

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -591,8 +591,6 @@ class TestClient(unittest.TestCase):
         wordlen, usrdata = self.client._prepare_as_read_area(area, amount)
         pusrdata = ctypes.byref(usrdata)
         res = self.client.as_read_area(area, dbnumber, start, amount, wordlen, pusrdata)
-        if res != 0:
-            self.fail(f"Client failed: --  as_read_area command returned {res}")
         self.client.wait_as_completion(1000)
         self.assertEqual(bytearray(usrdata), data)
 
@@ -604,8 +602,6 @@ class TestClient(unittest.TestCase):
         wordlen, usrdata = self.client._prepare_as_read_area(area, amount)
         pusrdata = ctypes.byref(usrdata)
         res = self.client.as_read_area(area, dbnumber, start, amount, wordlen, pusrdata)
-        if res != 0:
-            self.fail(f"Client failed: --  as_read_area command returned {res}")
         self.client.wait_as_completion(1000)
         self.assertEqual(bytearray(usrdata), data)
 
@@ -617,8 +613,6 @@ class TestClient(unittest.TestCase):
         wordlen, usrdata = self.client._prepare_as_read_area(area, amount)
         pusrdata = ctypes.byref(usrdata)
         res = self.client.as_read_area(area, dbnumber, start, amount, wordlen, pusrdata)
-        if res != 0:
-            self.fail(f"Client failed: --  as_read_area command returned {res}")
         self.client.wait_as_completion(1000)
         self.assertEqual(bytearray(usrdata), data)
 
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -513,7 +513,8 @@ class TestClient(unittest.TestCase):
         self.client.write_area(area, dbnumber, start, data)
         # start as_request and test
         wordlen, usrdata = self.client._prepare_as_read_area(area, size)
-        res = self.client.as_read_area(area, dbnumber, start, size, wordlen, usrdata)
+        pusrdata = ctypes.byref(usrdata)
+        res = self.client.as_read_area(area, dbnumber, start, size, wordlen, pusrdata)
         if res != 0:
             self.fail(f"Async read area failed before checking -- Error {res}")
         self.client.wait_as_completion(timeout)
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -515,8 +515,6 @@ class TestClient(unittest.TestCase):
         wordlen, usrdata = self.client._prepare_as_read_area(area, size)
         pusrdata = ctypes.byref(usrdata)
         res = self.client.as_read_area(area, dbnumber, start, size, wordlen, pusrdata)
-        if res != 0:
-            self.fail(f"Async read area failed before checking -- Error {res}")
         self.client.wait_as_completion(timeout)
         self.assertEqual(bytearray(usrdata), data)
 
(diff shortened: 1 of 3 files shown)


swamper123:
I may elaborate a bit:

- There have been SEGFAULTS and SIGABRT, because the handling was not correct in some cases
- test_wait_as_completion_pass got more timeout time, since these are ms not sec. This may cause a timeout, even unwanted.
- the value of wait_as_completion_pass may not be correct, since as_read_area was not checked correctly.
- as_read_area shall not return the value of the read process, it should return the result of the Cli_AsReadArea method (which is, if AsReadArea was successfull or not, since the pointer to a variable is passed)
- Waitloop in client_async was fixed with that as well

Since then, I can execute test_client.py multiple times without any issues happening.

test/test_client.py
```diff
@@ -572,30 +588,39 @@ def test_as_read_area(self):
...
         self.client.write_area(area, dbnumber, start, data)
-        res = self.client.as_read_area(area, dbnumber, start, amount)
-        self.client.wait_as_completion(10)
-        res2 = bytearray(res)
-        self.assertEqual(res2, data)
+        wordlen, usrdata = self.client._prepare_as_read_area(area, amount)
+        pusrdata = ctypes.byref(usrdata)
+        res = self.client.as_read_area(area, dbnumber, start, amount, wordlen, pusrdata)
+        if res != 0:
+            self.fail(f"Client failed: --  as_read_area command returned {res}")
```
gijzelaerr, line 595: just do a `self.assertEqual(res, 0)`, no need to a custom error message

test/test_client.py
```diff
@@ -572,30 +588,39 @@ def test_as_read_area(self):
...
```
gijzelaerr, line 608: just do a `self.assertEqual(res, 0)`, no need to a custom error message

test/test_client.py
```diff
@@ -572,30 +588,39 @@ def test_as_read_area(self):
...
```
gijzelaerr, line 621: just do a `self.assertEqual(res, 0)`, no need to a custom error message

test/test_client.py
gijzelaerr: also, doesn't make check_result() make sure it is always 0? so no      nee dfor the double check.

gijzelaerr (changes requested):

test/test_client.py
swamper123: I haven't looked at that since a while, but I guess this is a leftover while testing. I received sometimes values, which haven't been cought by check_result(). I guess the assert statement in error_text() needs to be replaced by `if context in ("client", "server", "partner")`, since assert can be ignored while running in process. And if there is something realy strange received, log that error.

But you are right, that this shouldn't be checked here again.

test/test_client.py
```diff
@@ -604,8 +623,11 @@ def test_as_write_area(self):
         size = 1
         start = 1
         data = bytearray(b'\x11')
-        self.client.as_write_area(area, dbnumber, start, data)
-        self.client.wait_as_completion(10)
+        wordlen, cdata = self.client._prepare_as_write_area(area, data)
+        res = self.client.as_write_area(area, dbnumber, start, size, wordlen, cdata)
+        if res != 0:
```
gijzelaerr, line 628: still some leftovers here

test/test_client.py
```diff
@@ -614,8 +636,11 @@ def test_as_write_area(self):
         dbnumber = 0
         size = 2
         timer = bytearray(b'\x12\x00')
-        self.client.as_write_area(area, dbnumber, start, timer)
-        self.client.wait_as_completion(10)
+        wordlen, cdata = self.client._prepare_as_write_area(area, timer)
+        res = self.client.as_write_area(area, dbnumber, start, size, wordlen, cdata)
+        if res != 0:
```
gijzelaerr, line 641: still some leftovers here

test/test_client_async.py
```diff
@@ -43,15 +43,20 @@ async def test_wait_loop(testclient):
...
+    # Test read_area with a DB
+    area = snap7.types.areas.DB
+    dbnumber = 1
+    data = bytearray(b'\x11')
+    testclient.write_area(area, dbnumber, start, data)
     # Execute test
-    p_data = testclient.as_db_read(db, start, size)
+    wordlen, usrdata = testclient._prepare_as_read_area(snap7.types.S7AreaDB, amount)
+    res = testclient.as_read_area(area, dbnumber, start, amount, wordlen, usrdata)
+    if res != 0:
```
gijzelaerr, line 56: still some leftovers here

test/test_client.py
```diff
@@ -624,8 +649,11 @@ def test_as_write_area(self):
         dbnumber = 0
         size = 2
         timer = bytearray(b'\x13\x00')
-        self.client.as_write_area(area, dbnumber, start, timer)
-        self.client.wait_as_completion(10)
+        wordlen, cdata = self.client._prepare_as_write_area(area, timer)
+        res = self.client.as_write_area(area, dbnumber, start, size, wordlen, cdata)
+        if res != 0:
```
gijzelaerr, line 654: still some leftovers here

test/test_client.py
```diff
@@ -542,18 +553,23 @@ def test_check_as_completion(self, timeout=5):
...
-        # Execute test
-        p_data = self.client.as_db_read(db, start, size)
+        start = 1
+        self.client.write_area(area, db, start, data)
+
+        # start as_request and test
+        wordlen, cdata = self.client._prepare_as_read_area(area, size)
+        pcdata = ctypes.byref(cdata)
+        res = self.client.as_read_area(area, db, start, size, wordlen, pcdata)
+        if res != 0:
```
gijzelaerr, line 567: still some leftovers here

test/test_client.py
```diff
@@ -524,16 +528,23 @@ def test_wait_as_completion_timeouted(self, timeout=0, tries=500):
...
         start = 1
         data = bytearray(size)
+        wordlen, data = self.client._prepare_as_read_area(area, size)
+        pdata = ctypes.byref(data)
         self.client.write_area(area, dbnumber, start, data)
         # start as_request and wait for zero seconds to try trigger timeout
         for i in range(tries):
-            self.client.as_db_read(dbnumber, start, size)
+            res = self.client.as_read_area(area, dbnumber, start, size, wordlen, pdata)
+            if res != 0:
```
gijzelaerr, line 537: still some leftovers here

test/test_client.py
```diff
@@ -512,9 +512,13 @@ def test_wait_as_completion_pass(self, timeout=10):
         data = bytearray(size)
         self.client.write_area(area, dbnumber, start, data)
         # start as_request and test
-        p_data = self.client.as_db_read(dbnumber, start, size)
+        wordlen, usrdata = self.client._prepare_as_read_area(area, size)
+        pusrdata = ctypes.byref(usrdata)
+        res = self.client.as_read_area(area, dbnumber, start, size, wordlen, pusrdata)
+        if res != 0:
```
gijzelaerr, line 518: still some leftovers here

gijzelaerr (changes requested):

PR #212, opened 2020-11-26
Change check to almostequal to avoid rare test_fails about time

### LINKED ISSUE #208: Runtime Issue in test_get_plc_time
It is possible, that in an unpractical moment `test_get_plc_time` fails, because `second` is different.
We should accept 1 second difference (should work with assertAlmostEqual?), since Python should be finished with the execution in that second.

swamper123:
Fix #208

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -4,7 +4,7 @@ import logging
 import struct
 import time
 import unittest
-from datetime import datetime
+from datetime import datetime, timedelta
 from multiprocessing import Process
 from os import kill
 from unittest import mock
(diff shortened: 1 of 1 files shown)


swamper123:
> 
> System ERROR:
> The system cannot find the file specified.
> Copy-Item: D:\a\_temp\45d27c9b-f308-4557-a922-c17d6f4fb47a.ps1:4
> Line |
>    4 |  Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll snap7\.
>      |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>      | Cannot find path 'D:\a\python-snap7\python-snap7\snap7-full-1.4.2\release\Windows\Win64\snap7.dll'
>      | because it does not exist.
> 
> Error: Process completed with exit code 1.

Looks like an error while downloading, but I guess it wouldn't change the result that the windows part won't pass (like the others).

Issue #203, opened 2020-10-28
Tests fails on windows


gijzelaerr:
```Run pytest.exe
============================= test session starts =============================
platform win32 -- Python 3.7.9, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: D:\a\python-snap7\python-snap7
collected 172 items

test\test_client.py .s.sss.ssssssssss........................s.......... [ 30%]
...........................................                              [ 55%]
test\test_client_async.py Entering debug mode. Use h or ? for help. 

At D:\a\_temp\8ec68c21-4c58-41b6-9035-e080ce31506e.ps1:3 char:5
+ if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXIT …
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[DBG]: PS D:\a\python-snap7\python-snap7>> 
Error: Process completed with exit code 1.
```

https://github.com/gijzelaerr/python-snap7/runs/1319367971?check_suite_focus=true

swamper123:
The client_async.py and test_client_async stuff will be revised in near future anyway, when the as_* stuff in client.py is working and accepted. There are known problems with asyncio <=3.7 and windows machines. I will have that in my mind while working on the asyncio stuff later.

swamper123:
If this is hinder your test here, you can delete that test folder and the client_async.py as well.
They both has to be revised anyway, when the as_* stuff in client.py is working and accepted. I got a branch for that purpose based on [61c970c0](https://github.com/gijzelaerr/python-snap7/commit/61c970c09e5ce1a2c30b1ed3e71fe16f19ffb527).

swamper123:
The same thing happens in test_partner as well:

https://github.com/gijzelaerr/python-snap7/pull/200/checks?check_run_id=1321783239

swamper123:
Why does test_partner doesn't need a tearDown to stop Partner()? 
Could this cause some trouble here?

swamper123:
So it realy seems to be a "Server in Background" thing as mentioned in #206 

In the experimental draft #205 the windwos tests are now running ... but now the python tests fail. 3.6 and 3.7 is fine.
Python 3.8 is just crashing without any hint and Python3.9 has an assertion error, I can't understand.

[...]

PR #213, opened 2020-11-30
Fix windows_test action

nikteliy:
1.  wget saves the archive with an incorrect name. I set output filename to fix "The system cannot find the file specified" error. 
2.  os.kill(pid, 1) on Windows kills a process with exit code 1, which may be the cause of issue #203 . It looks like the .terminate() method kills the `mainloop` process correctly.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,10 +15,9 @@ jobs:
         run: choco install 7zip wget python
       - name: Get snap7
         run: |
-          wget.exe --content-disposition -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
+          wget.exe -O snap7-full-1.4.2.7z -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
           7z x snap7-full-1.4.2.7z
-          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll snap7\.
-          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll C:\Windows\System32\.
+          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll C:\Windows\System32\
       - name: Upgrade pip and wheel for all pythons
         run:  python.exe -m pip install --upgrade pip wheel
       - name: Install python libraries
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,9 +15,10 @@ jobs:
         run: choco install 7zip wget python
       - name: Get snap7
         run: |
-          wget.exe -O snap7-full-1.4.2.7z -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
+          wget.exe -O snap7-full-1.4.2.7z --content-disposition -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
           7z x snap7-full-1.4.2.7z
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -15,10 +15,9 @@ jobs:
         run: choco install 7zip wget python
       - name: Get snap7
         run: |
-          wget.exe --content-disposition -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
+          wget.exe -O snap7-full-1.4.2.7z -c https://sourceforge.net/projects/snap7/files/1.4.2/snap7-full-1.4.2.7z/download
```
gijzelaerr, line 18: if I look at the latest master build https://github.com/gijzelaerr/python-snap7/runs/1340797153?check_suite_focus=true

the file is saved with the right name. Are you sure you are using `wget.exe` and not the powershell alias? also the extraction works fine:

```
2020-11-02 07:41:02 (4.99 MB/s) - 'snap7-full-1.4.2.7z' saved [20591252/20591252]


7-Zip 19.00 (x64) : Copyright (c) 1999-2018 Igor Pavlov : 2019-02-21

Scanning the drive for archives:
1 file, 20591252 bytes (20 MiB)

Extracting archive: snap7-full-1.4.2.7z
```

.github/workflows/windows.yml
```diff
@@ -15,10 +15,9 @@ jobs:
...
           7z x snap7-full-1.4.2.7z
-          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll snap7\.
-          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll C:\Windows\System32\.
+          Copy-Item snap7-full-1.4.2\release\Windows\Win64\snap7.dll C:\Windows\System32\
```
gijzelaerr, line 20: we copy the dll into the folder to prepare for the binary wheel.

test/test_client.py
```diff
@@ -37,7 +37,7 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
-        kill(cls.process.pid, 1)
+        cls.process.terminate()
```
gijzelaerr, line 40: interesting, that could be it. thanks :)

gijzelaerr (changes requested):

.github/workflows/windows.yml
nikteliy: Are you sure it works correctly? If snap7 folder doesn't exist then a file with that name will be created. Sorry, I probably don't fully understand.
copy-item-test

.github/workflows/windows.yml
nikteliy: I noticed an error at this build: https://github.com/gijzelaerr/python-snap7/runs/1457872062?check_suite_focus=true . And I checked it on another computer. It seems to me Sourceforge servers started sending an invalid file name.
wgetwtf

gijzelaerr: ok then it makes sense. note that I added  `--content-disposition `  to make sure wget follows the redirects, but the documentation is a bit minimal on this.

.github/workflows/windows.yml
gijzelaerr: this is executed in the python-snap7 folder, right? so a `snap7` module should exist.

nikteliy: Now I get it(and feel stupid). Thanks for the explanation. I'll move it back.

.github/workflows/windows.yml
nikteliy: You're right, better keep it. I've checked we can have both options.

.github/workflows/windows.yml
gijzelaerr: no worries, you are still being very useful :)

Issue #196, opened 2020-09-25
Feature request: ReadSZL()


NicklasMD:
Hi I didn't know where the right place where to ask for this?
The Snap7 has a function to read SZL values or lists of values, I would like that also in the Python-Snap7
Thanks for all your work.

**Example from the Snap7 documentation:**
Cli_ReadSZL
**Description**
Reads a partial list of given ID and INDEX.
**Declaration**
int Cli_ReadSZL( S7Object Client, int ID, int Index,
TS7SZL *pUsrData, int *Size);
function Cli_ReadSZL(Client : S7Object; ID, Index : integer;
pUsrData : PS7SZL; var Size : integer) : integer;
**Parameters**
Type Dir.
Client Native Integer In
The handle as return value of
Cli_Create(), passed by value.
ID integer 32 In List ID
Index integer 32 In List Index
pUsrData Pointer to struct in Address of the user buffer
size Pointer to integer 32 In Buffer size available
Out Bytes rea

swamper123:
What is an SZL btw? Never heared of that thing in an SPS context before (or can't remember).

This feature is actually not implemented, but feel free to make a PR.
I would recommened that you check other functions/methods and adapt their "wrapping behaviour".

NicklasMD:
The SZL's are used to get diagnostic data out of the PLC
For example status on the fieldbus, diagnostic buffer, status LED's etc. so I think it's a quite useful feature.

I'm not sure I'm competent enough to make the feature myself...

swamper123:
Found out SZL means "SystemZustandsListe", so like you said a diagnostic datas.

I thought I may never contribute to this repo as well, but after a while I am pretty into it in any kind of way. 😄  
If you inspect other client requests of their structure, you'll see that they are pretty similar and you see where `self.library....` is used, that this is the actual C command.

Otherwise you may have to wait a longer time until somebody implements it or you have to make PLC workarounds.

lautarodapin:
I really would love that two, I was looking for something like that for checking if devices where connected 
Didn't know that szl was the option
I will check if i can't get to something

PR #217, opened 2020-12-01
add the readszl() function #196

nikteliy:
Feature request #196 

* add the function
* add types for a partial list and for the header of a partial list
* add tests

Code has been tested with 1200(151-214-1hg31-0xb0) and ET200S(151-8ab01-0ab0).
I'm not sure about the return type. Is it okay to return a ctypes.Structure? Or is it better to convert it to dict, for example?

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -836,7 +836,7 @@ class Client:
         # Cli_ReadMultiVars
         raise NotImplementedError
 
-    def readszl(self, ssl_id: int, index: int=0x0000) -> S7SZL:
+    def readszl(self, ssl_id: int, index: int = 0x0000) -> S7SZL:
         # Cli_ReadSZL
         data = S7SZL()
         size = c_int(sizeof(data))
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -11,8 +11,7 @@ import snap7
 from snap7.common import check_error, load_library, ipv4
 from snap7.exceptions import Snap7Exception
 from snap7.types import S7Object, buffer_type, buffer_size, BlocksList
-from snap7.types import TS7BlockInfo, param_types, cpu_statuses
-from snap7.types import SystemStatusListHeader, SystemStatusList, S7SZLHeader, S7SZL
+from snap7.types import TS7BlockInfo, param_types, cpu_statuses, S7SZL
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


gijzelaerr:
Thank you for your contribution! Looking good and good question. I think I would use a named tuple unless you have a good argument against? I prefer to shield the user from the low-level details in a pythonic way.

swamper123:
@gijzelaerr Intuitively I would make a dict out of it, since the structure invites to make one out of it (a tuple with a name and a value in a list).

gijzelaerr:
I don't mean a tuple but a named tuple:

https://docs.python.org/3/library/collections.html#collections.namedtuple

I would argue this is the python equivalent of a struct since it preserves order, have a fixed set of properties and is prettier to use since it has property accessors. Since Python 3.9 dictionaries have an order also, but we support older pythons.

swamper123:
Oh, I haven't known that such a thing exists...well in that case, this would be the better choice. Thanks for the info!

nikteliy:
What if add [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple) in types.py and use it as return type? We'll have typed namedtuple similar to dataclass.
```python
import typing

class SystemStatusListHeader(typing.NamedTuple):
    length_dr: int
    number_dr: int


class SystemStatusList(typing.NamedTuple):
    header: SystemStatusListHeader
    data: bytes
```

gijzelaerr:
yes, I think that is perfect. Indeed, dataclass and namedtuple became more similar over time, but I think a named tuple is a bit leaner but I might be wrong.

thanks!

gijzelaerr:
I now realise that this breaks consistency with the rest of the code. What do you think works better from a developer perspective, the structs or the named tuples @nikteliy ?  If it improves the usability maybe we should aim to move everything to named tuples in ht belong run if it is worth the effort, otherwise, it might be better to stick with keeping the structs.

nikteliy:
> this breaks consistency with the rest of the code

In which way?
I don't see how it's possible to replace structures with tuples. They are at different levels of the application for me, as in the diagram.
pic1

nikteliy:
I think I should revert the last commit. Functions should behave in the same way. Named tuples are not useful enough to break backward compatibility. And this pr is not about tuples. What do you think @gijzelaerr ?

gijzelaerr:
Hehe, I love your drawing though :) 

I agree that shielding the scary low-level things is ideal, but this is not how python-snap7 has been set up initially. So maybe, for now, let's keep things consistent and maybe, at some point, we will switch to a more pythonic wrapping. Would be nice to get a 1.0 out the door someday, was planning to release that when the full API is implemented. for 2.0 I would like to switch to CFFI as a backend since that will make it possible/easy to create a binary wheel, eliminating many of the issues people experience while setting up python-snap7. 2.0 would be a good moment to switch to the pythonic type wrapping.

With the current pace of development that could be years away though :) I maintain this repo in my spare time, so unless someone pays me to rewrite the library I only accept PR's by developers like yourself and try to steer the project in the right direction.

thanks!

nikteliy:
> let's keep things consistent

"If it's works, don't touch it"  is my motto :)

> Would be nice to get a 1.0 out the door someday

Where can I find a list of tasks ordered by priority?

gijzelaerr:
The issue tracker :)

https://github.com/gijzelaerr/python-snap7/issues

although it is not really ordered. I want to make a 1.0 release if all bugs are resolved and all functions are implemented. Maybe the type hinting should be moved to 2.0 if the return types are going to change.

nikteliy:
> all bugs are resolved and all functions are implemented

It's a lot of work. I'll see what I can do.

gijzelaerr (approved):

PR #219, opened 2020-12-03
Add return Value type annotation

swamper123:


snap7/client.py
```diff
@@ -31,25 +32,25 @@ class Client:
     A snap7 client
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
```
gijzelaerr, line 35: I think `None` is the default return type, so you can save yourself the time to add it :)

snap7/client.py
```diff
@@ -209,7 +210,7 @@ def upload(self, block_num):
         return bytearray(_buffer)
 
     @error_wrap
-    def download(self, data, block_num=-1):
+    def download(self, data, block_num=-1) -> Any:
```
gijzelaerr, line 213: Why do you use Any here?

snap7/common.py
```diff
@@ -66,7 +66,7 @@ def check_error(code, context="client"):
         raise Snap7Exception(error)
 
 
-def error_text(error, context="client"):
+def error_text(error, context="client") -> Array:
```
gijzelaerr, line 69: Array? why not just a str

snap7/logo.py
```diff
@@ -53,7 +54,7 @@ def destroy(self):
         logger.info("destroying snap7 client")
         return self.library.Cli_Destroy(byref(self.pointer))
 
-    def disconnect(self):
+    def disconnect(self) -> Any:
```
gijzelaerr, line 57: We should not use Any i think.

snap7/server.py
```diff
@@ -42,10 +43,10 @@ def __init__(self, log=True):
         if log:
             self._set_log_callback()
 
-    def __del__(self):
+    def __del__(self) -> None:
         self.destroy()
 
-    def event_text(self, event):
+    def event_text(self, event) -> ctypes.Array:
```
gijzelaerr, line 49: why not jus tstr?

gijzelaerr (changes requested):

swamper123:
@gijzelaerr Don't spend to much time making reviews at the moment, since there will be changes and "tests" be pushed. 
When this Draft is ready, I will turn it into a normal PR and then you can take it apart. 🙈

gijzelaerr:
I would only use `Any` if a function really can return an ambiguous type, or the number of types just becomes to complex to enumerate. But that is probably bad code anyway.

swamper123:
Does somebody have a clue, why Mypy has a problem with `get_cpu_state(self) -> str:` ?
The value status_string is a string and can't get None, because it would be raised as an error then. 😕 

As well `load_library()` doesn't behave the way as intended. `self.library` is `None` and so the annotations are getting wrong, since an self.library.here_partner_func attribute seems to be expected.

swamper123:
It seems to be a mypy problem itself.
There have been errors appearing when a function/method doesn't contain a `return None`. This could be solved by:` --no-warn-no-return` [source](https://mypy.readthedocs.io/en/stable/command_line.html?highlight=annotation%20return%20None#configuring-warnings)

How the situation looks like with the cdll data in mypy, I am still not sure.

nikteliy:
There is some a problem with "Windows tests" after my last commit. But I have no idea how I could break it. The tests take too long.

nikteliy:
@gijzelaerr  сould you cancel the test before github sues us for wasting computing resources?

gijzelaerr:
at some point the test will just timeout, don't worry.

cancelled.

swamper123:
I read through some PRs and Issues from mypy and it seems that 

> snap7/client.py:93: error: Incompatible types in assignment (expression has type "None", variable has type "str")

is caused by the redefinition from str to None in some cases. A Workaround may be `--allow-redefinition`.

The issues in partner.py is indeed a `load_library()` problem, which isn't loading the library (so that class attributes are available). 
@gijzelaerr Shall I undo the failing changes?

snap7/client.py
```diff
@@ -49,55 +50,55 @@ def create(self):
...
         """
         Retrieves CPU state from client
         """
         state = c_int(0)
         self._library.Cli_GetPlcStatus(self._pointer, byref(state))
-
         try:
             status_string = cpu_statuses[state.value]
         except KeyError:
             status_string = None
```
gijzelaerr, line 93: i think mypy infers  `status_string` can be both a string and a None following this logic, if you just raise the snap exception in this block directly it probably won't be confused anymore.

swamper123: Jep, that worked. :)

gijzelaerr:
there is now a conflict with your previous pr

swamper123:
Fresh rebased

gijzelaerr (approved):

PR #220, opened 2020-12-08
Replace multiple ifs with elifs in util.py

swamper123:
Just a small improvement, where I stumbled over.
Or are there multiple ifs by purpose?


gijzelaerr:
not really, but it does work, so not a major issue.

I think by now i would solve it with a dict since that is compacter and faster
https://github.com/MonetDBSolutions/MonetDBe-Python/blob/master/monetdbe/monetize.py#L113

PR #227, opened 2020-12-16
Add async_client

nikteliy:
- add async functions to async_client

- add tests

- remove async functions from client.py

- remove async tests from test.client.py

- add async tests to action

### DIFF
--- a/snap7/client_async.py
+++ b/snap7/client_async.py
@@ -3,36 +3,131 @@ Snap7 async client used for connection to a siemens7 server.
 """
 import asyncio
 import logging
-from ctypes import byref
+from ctypes import byref, Array
 
 from snap7.common import check_error
+from snap7.types import S7SZL
 from .client import Client
 
-logger = logging.getLogger(__name__)
-
-
-def error_wrap(func):
-    """Parses a s7 error code returned the decorated function."""
-    async def f(*args, **kw):
-        code = await func(*args, **kw)
-        check_error(code, context="client")
-
-    return f
+from concurrent.futures import ThreadPoolExecutor
+from functools import wraps, partial
+from typing import Union, Tuple, Awaitable
 
 
 class ClientAsync(Client):
-    """
-    This class expands the Client class with asyncio features for async s7comm requests.
-    """
-
-    def __init__(self):
+    def __init__(self, max_workers=1):
         super().__init__()
-        self.as_check = None
+        self.executor = ThreadPoolExecutor(max_workers=max_workers)
 
-    async def wait_loop(self, operation):
+    def sync_to_async(self, func):
+        @wraps(func)
+        async def wrapper(*args, loop=None, executor=None, **kwargs):
+            if loop is None:
+                loop = asyncio.get_event_loop()
+            pfunc = partial(func, *args, **kwargs)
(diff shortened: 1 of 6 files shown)


snap7/client_async.py
```diff
@@ -3,36 +3,131 @@
...
+    def sync_to_async(self, func):
+        @wraps(func)
+        async def wrapper(*args, loop=None, executor=None, **kwargs):
+            if loop is None:
+                loop = asyncio.get_event_loop()
+            pfunc = partial(func, *args, **kwargs)
+            return await loop.run_in_executor(self.executor, pfunc)
+        return wrapper
 
+    async def as_compress(self, time) -> Awaitable[int]:
```
swamper123, line 31: I had a longer discussion once with @gijzelaerr about the topic where to store snap7-as_* methods.
The result was, that we left them in client.
I think it is possible to overload them in a Coroutine object without any trouble.

snap7/client_async.py
```diff
@@ -3,36 +3,131 @@
...
-    async def f(*args, **kw):
-        code = await func(*args, **kw)
-        check_error(code, context="client")
+class ClientAsync(Client):
+    def __init__(self, max_workers=1):
+        super().__init__()
+        self.executor = ThreadPoolExecutor(max_workers=max_workers)
 
-    return f
+    def sync_to_async(self, func):
```
swamper123, line 22: That is a pretty smooth way to convert functions to coroutines tbh!
But in the end, aren't the Coroutines blocking other running Coroutines? Could be ProcessPoolExecutor be the better choice?

test/test_client_async.py
```diff
@@ -1,28 +1,33 @@
...
+                        b'\x85V\x1f2\xa1\x00*\x00\x00\x00\x00\x00\x02\x01\x0f\x05c\x00#\x00\x00\x00' \
+                        b'\x11\x04\x10\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01' \
+                        b'\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x01\x04\x00' \
+                        b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' \
+                        b'\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
 
-@pytest.fixture
-def testserver():
+
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
nikteliy: Good point. I'll fix it.

snap7/client_async.py
nikteliy: I'll do as you say guys. Both options are good for me.

snap7/client_async.py
nikteliy: As far as I know, I/O calls release GIL, so a thread should be enough. But maybe you're right. I'll look into it.

swamper123: I made a research and it seems that you are right about that. I/O bounded calls seem to be the exception in the whole Thread case,.

PR #228, opened 2020-12-17
smooth process kill

swamper123:
Just a test draft, pls. ignore
Will be removed afterwards


swamper123:
Same failing behaviour as before. 😭

PR #239, opened 2021-01-11
smaller changes & add set_as_callback

swamper123:
Add set_as_callback to most neccessarry points
Correct some docstrings
Replace bytes with bytearray
Remove unused imports
Remove wrong parameter in docstring
Rename value, since it is used as an import statement
Change exception message with better expression
Rebase on master

_as_check_loop shall just be an optional tool for tests, instead of the using of wait_as_completion().

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -307,7 +307,7 @@ class Client:
         logger.debug(f"blocks: {blocksList}")
         return blocksList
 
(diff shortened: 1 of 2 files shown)


gijzelaerr (approved):
looks good to me. @nikteliy do you want to review?

PR #241, opened 2021-01-14
add test_set_as_callback

nikteliy:


swamper123:
Best way to get around for tests atm, until we find the source of all evil in that case for Windowssystems.
Perhaps a point worth on the "Still a Case"-Milestone list 🕵️ ?!

Issue #133, opened 2019-10-05
Can't handle indented DB layout specification


ronkinoz:
Requesting enhancement to util's parse_specification() so it can handle indented DB layout specifications.

This just requires the addition of lstrip() to the line processing.

`
        if line and not line.lstrip().startswith('#'):
`

spreeker:
create a pull requests.

gijzelaerr:
please.

gijzelaerr:
addressed in https://github.com/gijzelaerr/python-snap7/pull/254

PR #135, opened 2019-10-15
Enhancement to support indented db specifications. Fixes #133

ronkinoz:
Reworked line processing in parse_specification() such that indented db specifications are now supported.

PR #254, opened 2021-02-15
Add support for indented DB layout specifications

swamper123:
Fix for #133


gijzelaerr:
that was also implemented here, but we didn't merge it since it is not unit tested:

https://github.com/gijzelaerr/python-snap7/pull/135/files

gijzelaerr (approved):

PR #255, opened 2021-02-27
Types and lib location feature

lautarodapin:
Three main changes

1. Added optional parameter in the init client class, to specify the dll file location for those who can't installed properly.
2. Utils DB, added area parameter for handling marks, inputs and outputs. TODO: tests
3. Types: added Enum class for the areas, this way we can handle the typing.


snap7/util.py
```diff
@@ -522,7 +536,7 @@ def get_bytearray(self) -> bytearray:
             return self._bytearray._bytearray
         return self._bytearray
 
-    def export(self):
+    def export(self) -> Dict:
```
gijzelaerr, line 539: This dict type is missing the key and  value types (probably something like `Dict[str, int]`, i cant see that directly.

gijzelaerr (changes requested):
thanks for your contribution! If you address my comments and fix the mypy and pycodestyle errors we can merge this PR.

snap7/types.py
```diff
@@ -51,14 +52,24 @@
 mkEvent = 0
 mkLog = 1
 
+
 # Area ID
+class Areas(Enum):
```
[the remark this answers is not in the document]
lautarodapin: ADict could be use for typing as well right? If so, no need to use the enum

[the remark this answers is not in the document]
lautarodapin: Aparently ADict isn't as easy to use for typing as Enum, but switching everything to enum is cause some problems in other parts, it will take some time.

lautarodapin:
@gijzelaerr 
I don't really know how to handle this errors. 
```
snap7/common.py:44: error: Cannot determine type of 'cdll'
snap7/util.py:657: error: Argument 4 to "set_bool" has incompatible type "Union[bool, str, int, float]"; expected "bool"
snap7/util.py:667: error: Argument 3 to "set_string" has incompatible type "Union[bool, str, int, float]"; expected "str"
snap7/util.py:673: error: Argument 3 to "set_dword" has incompatible type "Union[bool, str, int, float]"; expected "int"
snap7/util.py:676: error: Argument 3 to "set_dint" has incompatible type "Union[bool, str, int, float]"; expected "int"
snap7/util.py:679: error: Argument 3 to "set_int" has incompatible type "Union[bool, str, int, float]"; expected "int"
snap7/util.py:682: error: Argument 3 to "set_word" has incompatible type "Union[bool, str, int, float]"; expected "int"
snap7/util.py:685: error: Argument 3 to "set_usint" has incompatible type "Union[bool, str, int, float]"; expected "int"
snap7/util.py:691: error: Argument 3 to "set_usint" has incompatible type "Union[bool, str, int, float]"; expected "int"
```

lautarodapin (commented):
I will work in migrating everythin to enums, seems better for typing

lautarodapin (commented):
Ok

lautarodapin:
@gijzelaerr I think is ready.

snap7/common.py
```diff
@@ -1,14 +1,16 @@
...
 from ctypes.util import find_library
+from typing import Optional, Union
 
 from snap7.exceptions import Snap7Exception
 
 if platform.system() == 'Windows':
     from ctypes import windll as cdll  # type: ignore
 else:
-    from ctypes import cdll
+    from ctypes import cdll  # type: ignore
```
gijzelaerr, line 13: why does this require ignoring?

snap7/common.py
```diff
@@ -21,7 +23,7 @@ class ADict(dict):
     Accessing dict keys like an attribute.
     """
     __getattr__ = dict.__getitem__
-    __setattr__ = dict.__setitem__
+    __setattr__ = dict.__setitem__  # type: ignore
```
gijzelaerr, line 26: why does this require ignoring?

snap7/common.py
```diff
@@ -87,3 +90,9 @@ def error_text(error, context: str = "client") -> bytes:
     elif context == "partner":
         library.Par_ErrorText(error, text, len_)
     return text.value
+
+
+def find_locally(fname):
+    if os.path.isfile("./" + fname + ".dll"):
```
gijzelaerr, line 96: i think you can make this more robust by using `pathlib.Path(__file__) / 'snap7.dll'` or something similar (maybe syntax incorrect)

snap7/types.py
```diff
@@ -212,6 +237,16 @@ class S7DataItem(ctypes.Structure):
         ('pData', ctypes.POINTER(ctypes.c_uint8))
     ]
 
+    def __str__(self) -> str:
```
gijzelaerr, line 240: i dont think this should be a multiline representation. This will give problems if you use this string representation in logging for example.

gijzelaerr (changes requested):
Thanks! Looking mostly good, if you address my comments i will merge the PR.

snap7/common.py
lautarodapin: It throws a warning saying that it can't determine the type of cdll

snap7/common.py
lautarodapin: Same here, I will reproduce the error later and show it

snap7/common.py
lautarodapin: Ok

snap7/types.py
lautarodapin: Ok, i will switch to the <__class__ (attr=val)> repr

snap7/common.py
gijzelaerr: mypy? but that is strange, we were testing with mypy before also.

lautarodapin: Here is the error.
````
snap7/common.py:26: error: Incompatible types in assignment (expression has type "Callable[[Dict[_KT, _VT], _KT, _VT], None]", base class "object" defined the type as "Callable[[object, str, Any], None]")
```

snap7/common.py
lautarodapin: Well now it doesn't show the same typing error, so i will remove it.

lautarodapin: This error i'm having, that is hy the ignore.
````
(venv) PS C:\Users\ldapino\Documents\lautaro\python\python-snap7>mypy snap7 test
snap7\common.py:14: error: Incompatible import of "cdll" (imported name has type "LibraryLoader[CDLL]", local name has type "LibraryLoader[WinDLL]")
```

lautarodapin:
@gijzelaerr ready i think

gijzelaerr (approved):

PR #258, opened 2021-03-15
feature_add_build_workflow

nikteliy:
* add a workflow for building wheels
* add a workflow for testing packages from test.pypi
* add a function to find a shared library inside package
* skip test_set_as_callback for macos


.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
+name: build-and-publish
+on:
+  push:
+    branches: [master]
+  pull_request:
```
gijzelaerr, line 5: i think tags are missing here, so the pypi upload will never be triggered.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+    name: Build binary wheels for linux
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - arch: x86_64
+            makefile: x86_64_linux.mk
+          - arch: i686
+            makefile: i386_linux.mk
```
gijzelaerr, line 20: who still uses i386?!? :)

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+          do
+          dirname=$wheelname-dir
+          packagename=$(python3 -m wheel unpack $wheelname -d $dirname | sed -e 's/.*\/\(.*\)...OK/\1/')
+          mkdir -p $dirname/$packagename/snap7/lib
+          mv $dirname/$packagename/python_snap7.libs/libsnap7-*.so $dirname/$packagename/snap7/lib/libsnap7.so
+          rm -rf $dirname/$packagename/python_snap7.libs
+          rm $dirname/$packagename/snap7/__dummy__*.so
+          python3 -m wheel pack $dirname/$packagename -d ./
+          rm -rf $dirname
+          done
```
gijzelaerr, line 73: whow i'm surprised this works. Did you test it on a system without libsnap7 in the LD_LIBRARY_PATH?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+        id: snap7-archive
+        uses: actions/cache@v2
+        with:
+          path: snap7-full-1.4.2.7z
+          key: ${{ env.snap7-archive-url }}
+
+      - name: Get snap7
+        if: steps.snap7-archive.outputs.cache-hit != 'true'
+        run: |
+          wget -O snap7-full-1.4.2.7z --content-disposition -c ${{ env.snap7-archive-url }}
```
gijzelaerr, line 40: love it.

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+          - arch: win_amd64
+            dll_dir: Win64
+          - arch: win32
+            dll_dir: Win32
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install choco packages
+        run: choco install 7zip wget python
```
gijzelaerr, line 96: so you only build this for 3.9? Ideally, we do it for 3.6, 3.7 and 3.8 also...

I've managed to get multiple pythons installed with choco on windows:

https://github.com/MonetDBSolutions/MonetDBe-Python/blob/master/.github/workflows/windows.yml#L24

but it is not the most stable solution.

.github/workflows/test-pypi-packages.yml
```diff
@@ -0,0 +1,42 @@
+name: test pipy packages
```
gijzelaerr, line 1: very cool you made this also.

setup.py
```diff
@@ -11,10 +12,20 @@
     'doc': ['sphinx', 'sphinx_rtd_theme'],
 }
 
-
 def read(fname):
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
+ext_modules = [
+    Extension(
+        "snap7.__dummy__",
```
gijzelaerr, line 20: so this is still required to trigger the binary wheel creation? or is it for another reason?

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,241 @@
...
+          }
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: wheels
+          path: wheelhouse/*.whl
+
+  osx-build:
+    name: Build binary wheels for osx
+    runs-on: macos-10.15
```
gijzelaerr, line 143: we might want to use pyenv to install multiple pythons

https://github.com/MonetDBSolutions/MonetDBe-Python/blob/master/.github/workflows/osx.yml#L25

gijzelaerr (changes requested):

gijzelaerr:
@nikteliy you really did a great job, these GitHub action jobs could be a great example of how cross-platform wheels could and should be built! well done.

.github/workflows/build-and-publish.yml
nikteliy: This is for the 32bit version of the library. If we don't need it, it will make thing easier. Do we need 32-bit versions?

.github/workflows/build-and-publish.yml
nikteliy: I did. It works for me if the package is installed as usual, either in a local user directory or in a virtual environment. I will add tests with next commit.

.github/workflows/build-and-publish.yml
nikteliy: I build one package for all python versions. It looks like python_snap7-1.0**-py3**-none-win_amd64.whl. Pip considers this package to be suitable for any 3.* version of python in Windows x64.

setup.py
nikteliy: That's right, auditwheel cannot determine dependencies without this. https://github.com/pypa/auditwheel#limitations

.github/workflows/build-and-publish.yml
nikteliy: cibuildwheel action creates a docker container for each python version. So we don't have to.

.github/workflows/build-and-publish.yml
gijzelaerr: if it is trivial to keep support then we should keep it.

.github/workflows/build-and-publish.yml
gijzelaerr: you are right! I didn't realise, but since we don't include anything compiled to the python library, we don't need to build for every python!  And i guess it is the same for all platforms actually. awesome.

.github/workflows/test-pypi-packages.yml
nikteliy: Without tests, no one will believe it works. Even me :)

.github/build_scripts/setup.py
```diff
@@ -0,0 +1,58 @@
+import os
+
```
gijzelaerr, line 2: we are almost there!! 

why a separate setup.py file?

nikteliy: In order not to affect other build options. For example, building from source will not work correctly if an extension is added to the setup.py. It also allows us keeps everything we need for build in one place and it's makes workflow a little cleaner.

gijzelaerr: isn't there an environment field we can check for so we know we are building a binary wheel? I don't like the idea of code duplication.

gijzelaerr (approved):

PR #259, opened 2021-03-16
Load standard values in server in testcases

lautarodapin:
I created in the mainloop a DB0 that will contain the prefix values.
We have to standarize where eachone will be located.

~~Also TODO:
I had problems for testing the real values, it seems that in the process of loading into c_int8 there is some residual memory left, and when you assert equal is not the same.~~


snap7/server.py
```diff
@@ -293,14 +294,75 @@ def clear_events(self) -> int:
...
+    DBdata = (snap7.types.wordlen_to_ctypes[snap7.types.WordLen.Byte.value] * size)()
+    PAdata = (snap7.types.wordlen_to_ctypes[snap7.types.WordLen.Byte.value] * size)()
+    TMdata = (snap7.types.wordlen_to_ctypes[snap7.types.WordLen.Byte.value] * size)()
+    CTdata = (snap7.types.wordlen_to_ctypes[snap7.types.WordLen.Byte.value] * size)()
     server.register_area(snap7.types.srvAreaDB, 1, DBdata)
     server.register_area(snap7.types.srvAreaPA, 1, PAdata)
     server.register_area(snap7.types.srvAreaTM, 1, TMdata)
     server.register_area(snap7.types.srvAreaCT, 1, CTdata)
+
+    ba = bytearray(1000)
```
gijzelaerr, line 306: i think we should move this to a separate function, and make it optional to initialise these values.

gijzelaerr (changes requested):

snap7/server.py
lautarodapin: Agreed.
Also, how we should standarize the locations for each values? 
Example:
Booleans from byte 0-10?
Int from 10-20? 
and so on?

gijzelaerr: i have no strong opinion about this.

swamper123: I would say, a simple structure, like you mentioned, would be nice. I like such rows to ten, easier to read and we are not restricted to tight space managament here.

lautarodapin: Ok, I will arrange it.
About the REAL, I had a lot of problems where I could not fix it.
I wrote in the server mainloop the real 3.123456e38 but when i read it in the client side i receive 3.1234567891e38 (the values are not real)

lautarodapin: This is the error that i'm getting it with real values.
```py
 def test_read_real(self):
        data = self.client.db_read(0, 37, 4 * 9)
        self.assertAlmostEqual(get_real(data, 0), -3.402823e38)
> AssertionError: -3.4028230607370965e+38 != -3.402823e+38 within 7 places (6.073709657548248e+30 difference)
```
I write the floating value directly to the bytearray like this
```py
ba[37:41] = struct.pack(">f", -3.402823e38)
```

lautarodapin: It seems like the `get_real` is the problem, because if i compare the bytearray before using the utils function both are the same
 ```py
    def test_read_real(self):
        data = self.client.db_read(0, 37, 4 * 9)
        print(data[0:4])
        self.assertEqual(data[0:4], b'\xff\x7f\xff\xfd')
>       self.assertAlmostEqual(get_real(data, 0), -3.402823e38)
E       AssertionError: -3.4028230607370965e+38 != -3.402823e+38 within 7 places (6.073709657548248e+30 difference)

test/test_mainloop.py:103: AssertionError
------------------------------------------------- Captured stdout call -------------------------------------------------
bytearray(b'\xff\x7f\xff\xfd')
```

Comment (lautarodapin) [reply snap7/server.py:306]: It seems that for some of the numbers provided by @swamper123 (ex: -3.402823e38) the presition is lost in single presition and we get -3.4028229e+38
image
In the test before the difference between the real value, and the read is 0.00001% (nothing) 
How should i proceed?

Comment (swamper123) [reply snap7/server.py:306]: Different sources, different numbers, but all from the forum on the siemens homepage (*grumble grumble*). 
Since REAL is 32bit long, single presition seems to be the right answer.

Comment (lautarodapin) [reply snap7/server.py:306]: Yes single presition is correct, i think it is a problem of presition for that big number, I added an assert almost equal with a delta of 0.0001% 

And for the string, I didn't understand what you said about the first two bytes, but it seems that the first is the total length (n+2) and the second one is the actual length (n)
image

Comment (swamper123) [reply snap7/server.py:306]: STRING[254] is the same as STRING.
n is the number of chars in the strings.

So STRING[254] is 254+2 Bytes big -> max. 256 Bytes

Like in the picture, Byte0 contains max length of a string and Byte1 the actural length of the string. In Byte3 to Byte255 contain the actual ASCII character.

**Comment by lautarodapin** [comment]
@gijzelaerr  @swamper123  I think is ready to merge.

**Comment by gijzelaerr** [comment]
@lautarodapin I think (but I'm not sure) you rewrote the git history on your fork, which is causing the weird divergence in history. I'll merge these changes manually since they look good, but I recommend you do a fresh new clone of the repo and start from there for future PR's.

File: snap7/util.py [hunk 2]
```diff
@@ -563,118 +586,123 @@ def get_offset(self, byte_index: Union[str, int]) -> int:
...
 
         byte_index = self.get_offset(byte_index)
 
-        if _type.startswith('STRING'):
-            max_size = re.search(r'\d+', _type).group(0)
-            max_size = int(max_size)
-            return set_string(bytearray_, byte_index, value, max_size)
+        if type.startswith('STRING') and isinstance(value, str):
+            max_size = re.search(r'\d+', type)
+            if max_size is None:
```
swamper123, line 670: Not sure what the better handling would be here.
I would say, if it is just STRING, we use the default of 254. (Maybe for STRING[] as well with a warning?).
If it is STRING[123] we take the number.
If it is STRING[GiBbErIsCh] or the number is not in range between 1 and 254, raise an error.

For longer strings, an extra datatype exist, which we don't support atm.

snap7/util.py
```diff
@@ -563,118 +586,123 @@ def get_offset(self, byte_index: Union[str, int]) -> int:
...
+            return set_usint(bytearray_, byte_index, value)
+
+        elif type == 'SINT' and isinstance(value, int):
+            return set_sint(bytearray_, byte_index, value)
+
+        if type == 'USINT' and isinstance(value, int):
             return set_usint(bytearray_, byte_index, value)
 
-        elif _type == 'SINT':
+        if type == 'SINT' and isinstance(value, int):
```
swamper123, line 700: SINT and USINT are already listed. you can remoive these two (with just the if statement)

test/test_common.py
```diff
@@ -0,0 +1,37 @@
...
+    def tearDown(self):
+        self.file.unlink()
+
+    def test_find_locally(self):
+        file = find_locally(file_name_test.replace(".dll", ""))
+        self.assertEqual(file, str(self.BASE_DIR / file_name_test))
+
+
+if __name__ == '__main__':
+    unittest.main()
```
swamper123, line 37: New line missing

test/test_mainloop.py
```diff
@@ -0,0 +1,139 @@
+import ctypes
+import logging
+from multiprocessing.context import Process
+from os import get_inheritable
```
swamper123, line 4: Unused import

swamper123 (changes requested):
It seems flake8 is not raising failures...have a look on it in the next PRs.

PR #262, opened 2021-03-23
add support for read/write byte from bytearray (no test implemented)

Buckler89:
Add utils to read/write byte from/to bytearray.
No unit test implemented at the moment.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -135,6 +135,28 @@ def set_bool(bytearray_: bytearray, byte_index: int, bool_index: int, value: boo
         bytearray_[byte_index] -= index_value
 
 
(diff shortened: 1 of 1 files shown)


lautarodapin:
Question, why would you need to set or get a specific byte?

Buckler89:
> Question, why would you need to set or get a specific byte?

Hi, 
  i need this features just because the type Byte is one of the type used to develop the PLC software. This choice (of using the type Byte I mean) is of course, developer depended. For example in a lot of PLC software i'm working with (i'm not the developer of PLC software), there of variables of type byte i need to set/get. 
Are there any drawbacks that i am not considering?

gijzelaerr:
i'm happy to merge this PR once a unittest is added.

Issue #242, opened 2021-01-19
Improve doc strings of all functions


gijzelaerr:
preferably in napoleon format (google python doc style)

lautarodapin:
@gijzelaerr What documentation should I follow? 
[this](http://snap7.sourceforge.net/sharp7.html)  ?

gijzelaerr:
yep!

gijzelaerr:
This is implemented!

PR #269, opened 2021-04-06
Improve doc strings of all functions

lautarodapin:
#242 
I started to create the doc string, please let me know if I'm doing this corrently.

PD: i follow your suggestion, delete the fork and refork it.

### DIFF
--- a/snap7/common.py
+++ b/snap7/common.py
@@ -29,9 +29,11 @@ class ADict(dict):
 
 
 class Snap7Library:
-    """
-    Snap7 loader and encapsulator. We make this a singleton to make
-    sure the library is loaded only once.
+    """Snap7 loader and encapsulator. We make this a singleton to make
+        sure the library is loaded only once.
+
+    Attributes:
+        lib_location (str, Optional): full path to the `snap7.dll` file. Defaults to None
     """
     _instance = None
     lib_location: Optional[str]
@@ -44,6 +46,14 @@ class Snap7Library:
         return cls._instance
 
     def __init__(self, lib_location: Optional[str] = None):
+        """ Loads the snap7 library using ctypes cdll.
+
+        Args:
+            lib_location (str, Optional): full path to the `snap7.dll` file. Defaults to None
+
+        Raises:
+            Snap7Exception: if `lib_location` is not found.
+        """
         if self.cdll:  # type: ignore
             return
         self.lib_location = (lib_location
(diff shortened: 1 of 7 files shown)


snap7/common.py
```diff
@@ -29,9 +29,11 @@ class ADict(dict):
...
 
 class Snap7Library:
-    """
-    Snap7 loader and encapsulator. We make this a singleton to make
-    sure the library is loaded only once.
+    """Snap7 loader and encapsulator. We make this a singleton to make
+        sure the library is loaded only once.
+
+    Attributes:
+        lib_location (str, Optional): full path to the `snap7.dll` file. Defaults to None
```
gijzelaerr, line 36: If I'm not mistaken, you don't need to include the types in the docstrings since they are taken from type hinting already.

gijzelaerr (changes requested):

snap7/common.py
lautarodapin: You are right.
I was following [this](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) example.
But according to [sphinx documentation](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#type-annotations) is not necessary. I will remove them.

snap7/common.py
```diff
@@ -44,6 +46,14 @@ def __new__(cls, *args, **kwargs):
         return cls._instance
 
     def __init__(self, lib_location: Optional[str] = None):
+        """ Loads the snap7 library using ctypes cdll.
+
+        Args:
+            lib_location: full path to the `snap7.dll` file. Defaults to None
```
gijzelaerr, line 52: From the type information, one already knows the default value is none. I would add "optional".

snap7/common.py
```diff
@@ -29,9 +29,11 @@ class ADict(dict):
...
 
 class Snap7Library:
-    """
-    Snap7 loader and encapsulator. We make this a singleton to make
-    sure the library is loaded only once.
+    """Snap7 loader and encapsulator. We make this a singleton to make
+        sure the library is loaded only once.
+
+    Attributes:
+        lib_location: full path to the `snap7.dll` file. Defaults to None
```
gijzelaerr, line 36: From the type information, one already knows the default value is none. I would add "optional".

gijzelaerr (changes requested):

snap7/common.py
lautarodapin: In case of optionals with a default value?

gijzelaerr: optional implies a default of `None`

gijzelaerr:
looking awesome! thanks

PR #306, opened 2021-08-02
Fixed typo area doesn't exist areas does

Ofloo:
Typo in example file area doesn't exist I assume it's supposed to be areas


gijzelaerr:
it looks like you are issuing a PR from an older checkout or version.

Ofloo:
Made the checkout today about an hour ago ?

Ofloo:
And the link in the comments also doesn't exist anymore

# https://github.com/gijzelaerr/python-snap7/blob/master/snap7/snap7types.py shoud be # https://github.com/gijzelaerr/python-snap7/blob/master/snap7/types.py

Ofloo:
And i made mistake myself, .. it should in fact be Areas rather then areas because areas gets a type error while Areas doens't.

gijzelaerr (approved):

PR #308, opened 2021-09-07
add get_time and set_time in util

Yingliangzhe:
Hello,
I have added the new functions in util. Could someone take a look at it?

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -1100,8 +1100,7 @@ class DB_Row:
 
         raise ValueError
 
-    def set_value(self, byte_index: Union[str, int], type: str, value: Union[bool, str, int, float]) -> Union[
-        bytearray, None]:
+    def set_value(self, byte_index: Union[str, int], type: str, value: Union[bool, str, int, float]) -> Union[bytearray, None]:
         """Sets the value for a specific type in the specified byte index.
 
         Args:
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -614,7 +614,7 @@ def get_time(bytearray_: bytearray, byte_index: int) -> str:
     byte_str = data_bytearray.hex()
     val = int(byte_str, 16)
     if (val & (1 << (bits - 1))) != 0:
-        sign = -1 # if sign bit is set e.g., 8bit: 128-255
+        sign = -1  # if sign bit is set e.g., 8bit: 128-255
         val = val - (1 << bits)  # compute negative value
         val = val * sign
 
@@ -623,9 +623,8 @@ def get_time(bytearray_: bytearray, byte_index: int) -> str:
     minutes = seconds // 60
     hours = minutes // 60
     days = hours // 24
-    time = str(days * sign) + ":" + str(hours % 24) + ":" + str(minutes % 60) + \
-           ":" + str(seconds % 60) + "." + str(milli_seconds)
-    return time
+    time_str = str(days * sign) + ":" + str(hours % 24) + ":" + str(minutes % 60) + ":" + str(seconds % 60) + "." + str(milli_seconds)
+    return time_str
 
 
 def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytearray:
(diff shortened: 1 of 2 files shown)


gijzelaerr:
first impression is that it looks good, thanks. there is only an issue with code style, as you can see in the github actions build:

```
snap7/util.py:617:18: E261 at least two spaces before inline comment
snap7/util.py:627:12: E127 continuation line over-indented for visual indent
snap7/util.py:657:109: W504 line break after binary operator
snap7/util.py:667:1: E302 expected 2 blank lines, found 1
```

snap7/util.py
```diff
@@ -586,8 +586,85 @@ def get_dt(bytearray_: bytearray, byte_index: int) -> str:
...
+            byte_index: byte index from where to start writing.
+            time_string: time value in string
+
+        Examples:
+            >>> data = bytearray(4)
+            >>> snap7.util.set_dint(data, 0, '-22:3:57:28.192')
+            >>> data
+                bytearray(b'\x8d\xda\xaf\x00')
+        """
+    import re
```
swamper123, line 649: `re` is already imported at line 88. This line is not needed.

Yingliangzhe:
Hello  @gijzelaerr,
I have modified the code style and pushed it again, github is checking the code. But one check (build-and-test-wheels / testing wheels (windows-latest, 3.7) (pull_request) ) is still pending after 30 mins. Could you please have a look at it? 
Thank you.

snap7/util.py
```diff
@@ -586,8 +586,85 @@ def get_dt(bytearray_: bytearray, byte_index: int) -> str:
...
+        sign = -1  # if sign bit is set e.g., 8bit: 128-255
+        val = val - (1 << bits)  # compute negative value
+        val = val * sign
+
+    milli_seconds = val % 1000
+    seconds = val // 1000
+    minutes = seconds // 60
+    hours = minutes // 60
+    days = hours // 24
+    time_str = str(days * sign) + ":" + str(hours % 24) + ":" + str(minutes % 60) + ":" + str(seconds % 60) + "." + str(milli_seconds)
```
swamper123, line 626: This string concatination style still works, but is a bit outdated (but not wrong). An f-string equivalent would be sth. like this:

time_str = f"{days * sign!s}):{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}")

swamper123 (changes requested):
Some smaller style things in my opinion.
A question I would have is, what happens if sb. sets a bigger/false `time_string` greater than 4 Bytes. Would it break succesfuly?

gijzelaerr:
regarding the pending CI build, you can ignore that, it just hangs sometimes.

snap7/util.py
```diff
@@ -586,8 +586,85 @@ def get_dt(bytearray_: bytearray, byte_index: int) -> str:
...
+    data_list = re.split('[: .]', time_string)
+    days, hours, minutes, seconds, milli_seconds = [int(x) for x in data_list]
+    if days < 0:
+        sign = -1
+    time_int = ((days * sign * 3600 * 24 + (hours % 24) * 3600 + (minutes % 60) * 60 + seconds % 60) * 1000 + milli_seconds) * sign
+    if sign < 0:
+        time_int = (1 << bits) + time_int
+    formatstring = '{:0%ib}' % bits
+    byte_hex = hex(int(formatstring.format(time_int), 2)).split('x')[1]
+    bytes_array = bytes.fromhex(byte_hex)
```
nikteliy, line 661: Are these conversions necessary? Int has a method .to_bytes() that does the same. Or it might be better to use a struct.

snap7/util.py
```diff
@@ -586,8 +586,85 @@ def get_dt(bytearray_: bytearray, byte_index: int) -> str:
...
+            Lower posible value is T#-24D_20H_31M_23S_648MS(-2147483648).
+
+        Args:
+            bytearray_: buffer to write.
+            byte_index: byte index from where to start writing.
+            time_string: time value in string
+
+        Examples:
+            >>> data = bytearray(4)
+            >>> snap7.util.set_dint(data, 0, '-22:3:57:28.192')
```
nikteliy, line 645: Typo. Wrong method.

snap7/util.py
Yingliangzhe: > Some smaller style things in my opinion.
> A question I would have is, what happens if sb. sets a bigger/false `time_string` greater than 4 Bytes. Would it break succesfuly?

Good catch. I didn't notice that issue. I will modify this and make a new pull request.

snap7/util.py
Yingliangzhe: You approach is more elegant. Thank you.

Yingliangzhe:
This version has one more problem. If the time_string is '-0:0:0:0.1', the result bytearray is not 0xffffffff, but 0x00000001.

Issue #310, opened 2021-09-08
This version has one more problem. If the time_string is '-0:0:0:0.1', the result bytearray is not 0xffffffff, but 0x00000001.


swamper123:
This version has one more problem. If the time_string is '-0:0:0:0.1', the result bytearray is not 0xffffffff, but 0x00000001.

_Originally posted by @Yingliangzhe in https://github.com/gijzelaerr/python-snap7/issues/308#issuecomment-915228239_

PR #311, opened 2021-09-08
modification of TIME data type

Yingliangzhe:
I have modified the code little bit and added some tests to solve #310

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -650,7 +650,6 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
             >>> data
                 bytearray(b'\x8d\xda\xaf\x00')
         """
-    import re
     sign = 1
     bits = 32
     data_list = re.split('[: .]', time_string)
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -1100,8 +1100,7 @@ class DB_Row:
 
         raise ValueError
 
-    def set_value(self, byte_index: Union[str, int], type: str, value: Union[bool, str, int, float]) -> Union[
-        bytearray, None]:
+    def set_value(self, byte_index: Union[str, int], type: str, value: Union[bool, str, int, float]) -> Union[bytearray, None]:
         """Sets the value for a specific type in the specified byte index.
 
         Args:
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -623,7 +623,7 @@ def get_time(bytearray_: bytearray, byte_index: int) -> str:
     minutes = seconds // 60
     hours = minutes // 60
     days = hours // 24
-    time_str = str(days * sign) + ":" + str(hours % 24) + ":" + str(minutes % 60) + ":" + str(seconds % 60) + "." + str(milli_seconds)
+    time_str = f"{days * sign!s}:{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}"
     return time_str
 
 
@@ -642,7 +642,7 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
 
         Examples:
             >>> data = bytearray(4)
-            >>> snap7.util.set_dint(data, 0, '-22:3:57:28.192')
+            >>> snap7.util.set_time(data, 0, '-22:3:57:28.192')
             >>> data
                 bytearray(b'\x8d\xda\xaf\x00')
         """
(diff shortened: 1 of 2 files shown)


snap7/util.py
```diff
@@ -586,8 +586,90 @@ def get_dt(bytearray_: bytearray, byte_index: int) -> str:
...
+            byte_index: byte index from where to start writing.
+            time_string: time value in string
+
+        Examples:
+            >>> data = bytearray(4)
+            >>> snap7.util.set_time(data, 0, '-22:3:57:28.192')
+            >>> data
+                bytearray(b'\x8d\xda\xaf\x00')
+        """
+    import re
```
swamper123, line 653: This `import re` is still not needed ;)

snap7/util.py
```diff
@@ -642,25 +644,34 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
...
         Examples:
             >>> data = bytearray(4)
-            >>> snap7.util.set_dint(data, 0, '-22:3:57:28.192')
+
+            >>> snap7.util.set_time(data, 0, '-22:3:57:28.192')
+
             >>> data
                 bytearray(b'\x8d\xda\xaf\x00')
         """
     import re
```
swamper123, line 653: This import statement is still not needed, since it is already imported at the top of the file.

Yingliangzhe:
Could the errors from Mypy be ignored? :)
I could not understand, the marked position is wrong.

Yingliangzhe:
@gijzelaerr Could you please merge the new version to master? Thank you!

snap7/util.py
```diff
@@ -642,25 +644,35 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
...
     bits = 32
     data_list = re.split('[: .]', time_string)
-    days, hours, minutes, seconds, milli_seconds = [int(x) for x in data_list]
-    if days < 0:
+    days: str = data_list[0]
+    hours: int = int(data_list[1])
+    minutes: int = int(data_list[2])
+    seconds: int = int(data_list[3])
+    milli_seconds: int = int(data_list[4])
+    if re.split(r'(\d+)$', days)[0:2][0] == '-':
```
nikteliy, line 661: The same, but a little more readable, imho.
```suggestion
    if re.match('^-\d{1,2}$', days):
```

snap7/util.py
```diff
@@ -642,25 +644,35 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
...
-    byte_hex = hex(int(formatstring.format(time_int), 2)).split('x')[1]
-    bytes_array = bytes.fromhex(byte_hex)
-    bytearray_[byte_index:byte_index + 4] = bytes_array
-    return bytearray_
+    if abs(int(days)) <= 24 and hours <= 23 and minutes <= 59 and seconds <= 59 and milli_seconds <= 999:
+        if int(days) * sign == 24 and hours > 20 and minutes > 31 and seconds > 23 and milli_seconds > 647 or \
+                int(days) * sign == -24 and hours > 20 and minutes > 31 and seconds > 23 and milli_seconds > 648:
+            raise ValueError('time value out of range, please check the value interval')
+
+        time_int = ((int(days) * sign * 3600 * 24 + (hours % 24) * 3600 + (minutes % 60) * 60 + seconds % 60) * 1000 + milli_seconds) * sign
```
nikteliy, line 668: This is a good bulletproof check, but this code is not easy to read (at least for me). It may be better to calculate the result first and then check the range. It is less efficient, but more readable.
```suggestion
            time_int = ((int(days) * sign * 3600 * 24 + (hours % 24) * 3600 + (minutes % 60) * 60 + seconds % 60) * 1000 + milli_seconds) * sign
            if time_int < -2147483648 or time_int > 2147483647:
                 raise ValueError('time value out of range, please check the value interval')
```

snap7/util.py
```diff
@@ -642,25 +644,35 @@ def set_time(bytearray_: bytearray, byte_index: int, time_string: str) -> bytear
...
+        if sign < 0:
+            time_int = (1 << bits) + time_int
+        bytes_array = time_int.to_bytes(4, byteorder='big')
```
nikteliy, line 671: The sign check can be replaced by passing a parameter `signed=True` to the method `.to_bytes()`.
```suggestion
        bytes_array = time_int.to_bytes(4, byteorder='big', signed=True)
```

nikteliy (changes requested):

snap7/util.py
swamper123: This wouldn't solve problems like "negative seconds".

`time_string `must match with a strict pattern, before stuff is splitted. Otherwise stuff will break.
There could be other advantages, if the string was checked properly first, too.

Yingliangzhe: @swamper123  Yes, as I wrote this function, I also thought, in which format the `time_string` should be. 
I just use the format as the return value from `get_time()`.

swamper123: @Yingliangzhe The format is totally fine and I would have made the same way, if machines return it like that.
But unfortunatly, the Dr. Evils and "DAU"s (like we germans say :wink: ) like to break things. So inputs should be bulletproof. ^^

Yingliangzhe: @nikteliy this piece of code, you suggested, cannot cover the case, if I write '-24:**25**:30:23:193' to `time_string`. 
As intermediate result of `time_int`, it would be -2079023193 and still inside of the interval [-2147483648, 2147483647], but this time format should not be acceptable.

nikteliy: Yup, my mistake. I give up then :)

swamper123: `re.fullmatch(r"-?([2][0-4]|[1]\d|\d):([2][0-3]|[1]\d|\d):([1-5]\d|\d):([1-5]\d|\d):\d{1,3}.\d{1,3}", time_str)`

This should check all legit patterns (I needed a while to get in), but not if min/max is reached (because if the day is 24, hours shouldn't be greater than 20 .... well it's my first thought, but I got an idea. ^^

swamper123: @Yingliangzhe Could you verify if this is legit, even with min/max values?

```python
if re.fullmatch(r"(-?(2[0-3]|1?\d):(2[0-3]|1?\d|\d):([1-5]?\d):[0-6]?\d.\d{1,3})|"
                 r"(-24:(20|1?\d):(3[0-1]|[0-2]?\d):(2[0-3]|1?\d).(64[0-7]|6[0-3]\d|[0-5]\d{1,2}))|"
                 r"(24:(20|1?\d):(3[0-1]|[0-2]?\d):(2[0-3]|1?\d).(64[0-8]|6[0-3]\d|[0-5]\d{1,2}))"
```

swamper123: To clear this thing a bit up, each line is a case (normal, min, max):

1. (-)23 days : 0-23hours : 0-59minutes : 0-59seconds . 0-999 miliseconds
2. -24 days : 0-20hours: 0-31 minutes: 0-23 seconds. 0-647 milliseconds
3. 24 days : 0-20hours: 0-31 minutes: 0-23 seconds. 0-648 milliseconds

Yingliangzhe: Hello @swamper123,
sorry for the late answer, I have checked the solution you suggested. 
I don't think, it has no problem by interpreting the TIME data type. 
But, as far as I remember, the max value of TIME, the milliseconds should be 0 - 647, and the min value should be 0 - 648. I have changed the corresponding position.

Yingliangzhe:
Hello @gijzelaerr,
this version should fix the  #310. Could you please check and merge it to master? 
Thank you.

PR #313, opened 2021-09-27
Add byte type parsing to DB_Row API + set parsing case insensitive

LoicGRENON:
I added the BYTE type to the parsing to DB_Row API to avoid ValueError exception when parsing a specification with BYTE type value.
I also set the string value to upper case in order to do a case insensitive search for the type of the value when parsing the specification (e.g. : BYTE, Byte, byte, etc ... are all acceptable).

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -1069,6 +1069,7 @@ class DB_Row:
         """
 
         bytearray_ = self.get_bytearray()
+        type_ = type_.upper()
 
         if type_ == 'BOOL':
             byte_index, bool_index = str(byte_index).split('.')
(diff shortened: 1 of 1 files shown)


gijzelaerr:
thanks! The pycodestyle check fails though, can you fix that also?

LoicGRENON:
> thanks! The pycodestyle check fails though, can you fix that also?

Yeah, I see it failed. It's because I did the commit using the github web-editor and it automatically add indentation on each new lines ...
Unfortunately, I deleted the branch from my repo so I can't commit on it anymore.

Not sure what I could do.
Should I do another PR with corrected pycodestyle error ?
Can you accept this PR and then remove the indentation on the new line ?

swamper123:
I would say close this PR in that case, pull the master, create new branch, copy&pasta these view lines. This would be the cleanest way. :)

PR #314, opened 2021-09-28
Fix utils.get_time() for small negative values

LoicGRENON:
I noticed utils.get_time() doesn't work for small (below one day) negative values.
For example `T#-1ms` returns `0:0:0:1.0`.

With this fix it will returns `-0:0:0:1.0`.

BTW, i'm wondering if it wouldn't be better to return timedelta (and datetime object for all others time function as get_dt(), etc ...
Or at least to let the user the choice to get a string or datetime object, maybe with a function parameter.
What you think ?

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -645,7 +645,8 @@ def get_time(bytearray_: bytearray, byte_index: int) -> str:
     hours = minutes // 60
     days = hours // 24
 
-    time_str = f"{days * sign!s}:{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}"
+    sign_str = '' if sign >= 0 else '-'
+    time_str = f"{sign_str}{days!s}:{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}"
 
     return time_str
 


snap7/util.py
```diff
@@ -645,7 +645,8 @@ def get_time(bytearray_: bytearray, byte_index: int) -> str:
     hours = minutes // 60
     days = hours // 24
 
-    time_str = f"{days * sign!s}:{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}"
+    sign_str = '' if sign >= 0 else '-'
+    time_str = f"{sign_str}{days!s}:{hours % 24!s}:{minutes % 60}:{seconds % 60!s}.{milli_seconds!s}"
```
swamper123, line 649: there is a `!s` missing in minutes.

swamper123:
Is that caused by the value of the PLC or the wrong handling of the `get_time()` in general? I haven't got a device atm to test it.

swamper123:
We may need more test values for these kind of functions (extreme small (not negative), everything zero ...).

gijzelaerr:
i agree with the need for more tests to make sure this works properly.

LoicGRENON:
I added the missing !s to minutes (its currently missing on the current branch as well).

I'm currently doing my tests using a 1513-1 CPU.
I tested with T#0ms, T#1ms, T#1s, T#1m, T#1h, T#1d and their respective negative values.
I could write a test for this function as well.

LoicGRENON:
I can modify test.unit.py as this : 
`
    def test_get_time(self):
        test_values = [
            (0, '0:0:0:0.0'),
            (1, '0:0:0:0.1'),  # T#1MS
            (1000, '0:0:0:1.0'),  # T#1S
            (60000, '0:0:1:0.0'),  # T#1M
            (3600000, '0:1:0:0.0'),  # T#1H
            (86400000, '1:0:0:0.0'),  # T#1D
            (-0, '0:0:0:0.0'),
            (-1, '-0:0:0:0.1'),  # T#-1MS
            (-1000, '-0:0:0:1.0'),  # T#-1S
            (-60000, '-0:0:1:0.0'),  # T#-1M
            (-3600000, '-0:1:0:0.0'),  # T#-1H
            (-86400000, '-1:0:0:0.0'),  # T#-1D
        ]

        data = bytearray(4)
        for value_to_test, expected_value in test_values:
            data[:] = struct.pack(">i", value_to_test)
            self.assertEqual(snap7.util.get_time(data, 0), expected_value)
`

What you think ?

PR #316, opened 2021-10-04
Add support for read/write unsigned value from bytearray

LoicGRENON:
Hello,

I added get_uint/set_uint and get_udint/set_udint to add support for read/write for respectively uint (2bytes) and udint (4 bytes) values from bytearray.

get_usint/set_usint already exists.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -582,6 +582,57 @@ def set_dint(bytearray_: bytearray, byte_index: int, dint: int):
         bytearray_[byte_index + i] = b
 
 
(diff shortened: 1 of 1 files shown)


LoicGRENON:
BTW, I implemented this following the same way get_usint/set_usint was done.
I think it might be maybe clearer to just add "unsigned" parameter to get_int, get_dint, get_sint and their respective set functions and to set the struct format character accordingly rather same copying the same lines multiple time.

Do you want me to do such reformating of the functions ?

PR #342, opened 2022-01-18
Add py.typed marker file

mthuurne:
Currently `python-snap7` includes type annotations, but these are ignored when type-checking an application that uses this library. By including the `py.typed` marker file in the wheel, type checkers like `mypy` can check whether applications call `python-snap7` with the correct types.

https://www.python.org/dev/peps/pep-0561/#packaging-type-information


mthuurne:
The CI error is a timeout when installing tooling, so I guess it's an issue in the test infrastructure rather than a problem in the PR itself.

Issue #272, opened 2021-04-22
Drop Snap7Exception


gijzelaerr:
I think we should rely less on `Snap7Exception`, and replace it with common Python exceptions.

For example, if a user supplies the wrong type to a function we should raise a `TypeError`.

swamper123:
I guess the only left Snap7Exception would be here:
https://github.com/gijzelaerr/python-snap7/blob/ee7ad61a4a3ec297a7ce31e9223255c01801ed68/snap7/common.py#L78

Since this is the main interface to get informations from the underneath snap7?

lautarodapin:
> I guess the only left Snap7Exception would be here:
> https://github.com/gijzelaerr/python-snap7/blob/ee7ad61a4a3ec297a7ce31e9223255c01801ed68/snap7/common.py#L78
> 
> Since this is the main interface to get informations from the underneath snap7?

Is there anyreason to keep using the informations from the snap7 lib? Having different exceptions in python will be nice for handling errors during development with python-snap7.

swamper123:
Our interface with snap7 is pretty small for thrown errors and dealing with thrown errors from snap7 in Python is horrible if it "just happens". It is not like a nice "ValueError" is thrown, in case stuff underneath exploded, they are much more ugly.
This check_error is a nice way to seperate our lib with the snap7 one. Also we are able to understand what was thrown from snap7 and could track down problems in an easy way.

Discussable would be creating new Exceptions/add messages to typical python ones like ValueError and use these...but I believe using builtin ones is enough for our usecases

gijzelaerr:
there are a couple
```
λ  grep -R "raise Snap7Exception" snap7
snap7/common.py:            raise Snap7Exception(msg)
snap7/common.py:        raise Snap7Exception(error)
snap7/client.py:            raise Snap7Exception(f"The cpu state ({state.value}) is invalid")
snap7/client.py:            raise Snap7Exception("The blocktype parameter was invalid")
snap7/client.py:            raise Snap7Exception("The blocktype parameter was invalid")
snap7/client.py:            raise Snap7Exception("The parameter was invalid")
snap7/client.py:            raise Snap7Exception("The parameter was invalid")
snap7/client.py:            raise Snap7Exception("The blocktype parameter was invalid")
snap7/util.py:                raise Snap7Exception("Max size could not be determinate. re.search() returned None")
snap7/util.py:                raise Snap7Exception("Max size could not be determinate. re.search() returned None")
[...]

PR #343, opened 2022-01-18
fix #339 

### LINKED ISSUE #339: bug introduced in version 1.1
In version 1.1, following exception is thrown when executing "client.read_area(131, 0 , 1, 1)"
TypeError("unsupported operand type(s) for 'in': 'int' and 'EnumMeta'"
The same command runs without issue in version 0.11.

By step in debug, it is found the when testing whether or not an integer is in enumeration s7type.Areas causes the exception in client.py.
The enum "Areas" does not exist in version 0.11.

nikteliy:
fix #339
- Add integer -> Areas conversion
- Add exception
- Add tests

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -10,7 +10,7 @@ from typing import Union, Tuple, Optional, List
 
 import snap7
 from snap7.common import check_error, load_library, ipv4
-from snap7.exceptions import Snap7Exception
+from snap7.exceptions import Snap7Exception, Snap7InvalidArea
 from snap7.types import Areas, WordLen, S7Object, buffer_type, buffer_size, BlocksList, S7CpuInfo, S7DataItem, S7SZL, S7OrderCode, \
     S7Protection, S7SZLList, S7CpInfo
 from snap7.types import TS7BlockInfo, param_types, cpu_statuses
@@ -389,9 +389,11 @@ class Client:
             >>> buffer
             bytearray(b'\\x00\\x00')
         """
-        if area not in Areas:
-            raise ValueError(f"{area} is not implemented in snap7.types")
-        elif area == Areas.TM:
+        try:
+            area = Areas(area)
+        except ValueError:
+            raise Snap7InvalidArea(area)
+        if area == Areas.TM:
             wordlen = WordLen.Timer
         elif area == Areas.CT:
             wordlen = WordLen.Counter
@@ -425,6 +427,10 @@ class Client:
             >>> buffer = bytearray([0b00000001])
             >>> client.write_area(snap7.types.Areas.DB, 1, 10, buffer)  # Writes the bit 0 of the byte 10 from the DB number 1 to TRUE.
         """
+        try:
+            area = Areas(area)
+        except ValueError:
+            raise Snap7InvalidArea(area)
         if area == Areas.TM:
             wordlen = WordLen.Timer
         elif area == Areas.CT:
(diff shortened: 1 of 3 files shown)


snap7/client.py
```diff
@@ -389,9 +389,11 @@ def read_area(self, area: Areas, dbnumber: int, start: int, size: int) -> bytear
             >>> buffer
             bytearray(b'\\x00\\x00')
         """
-        if area not in Areas:
-            raise ValueError(f"{area} is not implemented in snap7.types")
-        elif area == Areas.TM:
+        try:
+            area = Areas(area)
```
gijzelaerr, line 393: This works if `area` is an `Area` object already? 

If we do want to add support for both `int` and `Area` types I think we should update the function signature to accept `area: Union[Area, int]` (or `area: Area | int` for py3.10 but we support older versions).

snap7/exceptions.py
```diff
@@ -1,5 +1,18 @@
...
+
+
 class Snap7Exception(Exception):
     """
     A Snap7 specific exception.
     """
     pass
+
+
+class Snap7InvalidArea(Snap7Exception):
```
gijzelaerr, line 11: I was trying to get rid of snap7 specific exceptions actually. In my opinion this is just a `ValueError`, and there is no need to create a new exception type.

test/test_client.py
```diff
@@ -1003,6 +1003,20 @@ def callback(FUsrPtr, JobOp, response):
...
+        start = 1
+        valid_area = 28
+        dbnumber = 0
+        data = bytearray(b'\x13\x35')
+        self.client.write_area(valid_area, dbnumber, start, data)
+        res = self.client.read_area(valid_area, dbnumber, start, amount)
+        self.assertEqual(data, bytearray(res))
+
+        invalid_area = 777
+        self.assertRaises(Snap7InvalidArea, self.client.read_area, invalid_area, dbnumber, start, amount)
```
gijzelaerr, line 1017: do we also test valid ints and Area as arguments for read_area?

snap7/client.py
nikteliy: >This works if area is an Area object already?

Yes, it works

>If we do want to add support for both int and Area 

We don't need to support `int` type and we shouldn't. Let's call it undocumented feature. 

Maybe i'm wrong, but in my opinion, `Areas` is a bunch of integer constants, with `area = Areas(area)` we just make sure the `area` value is one of them. If it works with integer - good for lazy people like me.

snap7/exceptions.py
nikteliy: I didn't want to write the same error message 4 times, so I created an exception that keeps the error message.  If  `ValueError: X is not a valid area` is enough, I could rid of "try/catch" too.

test/test_client.py
nikteliy: `test_write_area` and `test_read_area`  use the `Areas` type. `test_read_write_integer_area`  also checks a valid integer (line 1009).

snap7/client.py
gijzelaerr: I'm surprised mypy is not complaining since it is wrongly typed then? We run mypy on the test suite also, and it should raise an error then since int is not a valid argument for the function according to the typing.

I'm not sure if it is generally a good idea to ignore the type system; it helps find bugs while doing static code analysis. 

If the type checking would work correctly, a user is using mypy and this user wants to insert an int into this function he has to disable the type-checking of this function. So if we do add support for both types maybe it is best to be clear about that, and in the function check the type and act accordingly. The static code compiler probably understands that also.

snap7/exceptions.py
gijzelaerr: i think that is enough.

test/test_client.py
gijzelaerr: ah yes, sorry, i could have checked but was on my mobile phone.

snap7/client.py
gijzelaerr: Pycharm complains, i didn't figure out yet why mypy doesn't complain:

Schermafbeelding 2022-01-20 om 09 26 34

snap7/exceptions.py
gijzelaerr: btw, see discussion here https://github.com/gijzelaerr/python-snap7/issues/272

snap7/client.py
nikteliy: If mypy against me I have no choice but surrender. I was hoping that mypy regards IntEnum as a subclass of int, but it doesn't


image

PR #348, opened 2022-01-20
Rename test module to tests

gijzelaerr:


gijzelaerr:
this is too outdated by now, probably better to start over at some point. also not that important.

PR #357, opened 2022-03-15
add more getter methods for utils

swamper123:
fix shadow declarations like min and type
add S-1200 compatible getter functions
add missing, but not testable datatype getters

Tests will be added soon. This is a sneak-peak for now. Tested with a S1200 and supported datatypes.
Stuff, like LINT, can't be verified with a real device from me.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -411,14 +411,14 @@ def get_real(bytearray_: bytearray, byte_index: int) -> float:
     return real
 
 
-def set_string(bytearray_: bytearray, byte_index: int, value: str, max_size: int):
+def set_string(bytearray_: bytearray, byte_index: int, value: str, max_size: int  = 255):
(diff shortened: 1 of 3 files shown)


snap7/util.py
```diff
@@ -451,7 +451,7 @@ def set_string(bytearray_: bytearray, byte_index: int, value: str, max_size: int
         bytearray_[byte_index + 2 + r] = ord(' ')
 
 
-def get_string(bytearray_: bytearray, byte_index: int, max_size: int) -> str:
```
gijzelaerr, line 454: this breaks backwards compatibility, right?

swamper123: `max_size` is important for set/write methods, where you also have to be aware of the max. length.

If you get an already readed bytearray, the information is already in the bytearray.

Issue #366, opened 2022-06-07
util.set_string doesn't set max string size byte


swamper123:
In #291  was an error mentioned, which I have now as well.

https://github.com/gijzelaerr/python-snap7/blob/5b1fa247e6efab34b4af5fec77d5926293305d10/snap7/util.py#L414

`util.set_string` doesn't provide a max. char size (Byte 0) which leads to undefined behaviour on SPS side (seems that DB breaks partly).

Two changes are needed:

1) `bytearray_[byte_index + 1] = max_size` has to be added.
2) `     for r in range(i + 1, bytearray_[byte_index]):` has to be changed to `    for r in range(i + 1, bytearray_[byte_index] - 2):`, so that no out of range error appears. 

This would fix the wrong behaviour, PR will income soon.

PR #368, opened 2022-06-08
fix set_string

swamper123:
fix and add doc for set_string()
fix possible out of range error in set_string()
fix doc in get_string()
add non-ascii check and test

fixes #291  and #366


swamper123:
When #369 is merged, this branch will be rebased on it.
It fails, because 3.6 doesn't support `str.isascii()` function.

Issue #375, opened 2022-07-04
Support fixed size strings


vmsh0:
`python-snap7` supports length-prefixed strings, but it does not support fixed-length strings.

This is not such an issue when manually unpacking reads with a pattern such as

```
b = client.db_read(10, 0, 20)
varA = snap7.util.get_byte(b, 0)
varB = snap7.util.get_byte(b, 1)
varString = ""
for i in range(18):
   varString = varString + snap7.util.get_char(b, 2+i)
```

However, when using the `snap7.util.DB` class, this doesn't work, as `STRING`s are decoded with the expectation of being length-prefixed with no other option (but treating each character separately, which is not optimal).

If you're ok with it and there's the expectation that it will be pulled, I can provide a patch to implement a second set of string utility functions and support for the `DB` class, to deal specifically with fixed-length strings.

This could be implemented both as a new separate fstring type:
- new methods `get_fstring(...)` and `set_fstring(...)`
- a new type for `DB` layouts `FSTRING`

Or it could be implemented as a flag:
-  parameter `type` with possible values `prefixed` and `fixed` for `get_string(...)` and `set_string(...)` (taking care not to change the ordering and having `prefixed` as a default value, so that existing code is not broken)
- have a flag for the `STRING` type (such that it would look like `6   field_name   STRING[18]F`)

As mentioned, I can provide a patch for either approach. Please let me know whether you like any of those or if you have a different approach in mind to support this use case.

swamper123:
Hi,

sorry for the late response.

First: Helpfull PRs are always welcomed. :)

But tbh I still haven't understood 100%ly what you mean/your goal is. 

STRINGS are defined with a prefix (max. string length and actual string length), because Siemens said so in the specs. 
And if I say this value is a `STRING`, I would expect to be able to reuse it to communicate with a real PLC. Also the values in a PLC are stored as single characters as well ((array like). That's why we got a `byte_index ` parameter.

So why want you introduce/use a new `FSTRING` type, which is not defined in the specs?

vmsh0:
Thank you for the reply!

It's pretty common in industrial settings to find setups where strings
represented as arrays of chars, either 0-terminated or space-padded, as
opposed to using the actual length-prefixed string type.

So what I propose is adding a data type to this library to be able to parse
those representations as strings. Yes, that would be out of spec (as in
/not in the spec/, not /against the spec/), but it does represent an actual
use case.

We are actually doing that already on an internal fork of the library, and
[...]

PR #378, opened 2022-08-01
Add support to read/write space-padded fixed-length strings

vmsh0:
Following the discussion in issue #375, I'm offering this PR with the modifications we are using internally to support the datatype.

A recap of what this is, adapted from the commit message:

These strings are not part of the official s7comm specification, but
they are commonly found in real-world systems. They simply consist of a
byte array that contains characters forming a string, padded on the
right with spaces (ascii 32) when the string length is shorter than the
array.

This commit adds support for reading and writing, as well as usage in
tables (snap7.util.DB), and unit tests.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -411,6 +411,44 @@ def get_real(bytearray_: bytearray, byte_index: int) -> float:
     return real
 
 
+def set_fstring(bytearray_: bytearray, byte_index: int, value: str, max_length: int):
+    """Set space-padded fixed-length string value
+
+    Args:
+        bytearray_: buffer to write to.
+        byte_index: byte index to start writing from.
+        value: string to write.
+        max_length: maximum string length, i.e. the fixed size of the string.
+
+    Raises:
+        :obj:`TypeError`: if the `value` is not a :obj:`str`.
+        :obj:`ValueError`: if the length of the `value` is larger than the `max_size`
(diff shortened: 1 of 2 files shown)


vmsh0:
I kept this separate from #377 in case you only want the fix (and not this feature), but if you wish I can both changes on your last master and offer a single pull request with both things.

snap7/util.py
```diff
@@ -410,6 +410,44 @@ def get_real(bytearray_: bytearray, byte_index: int) -> float:
...
+        or 'value' contains non-ascii characters.
+
+    Examples:
+        >>> data = bytearray(20)
+        >>> snap7.util.set_fstring(data, 0, "hello world", 15)
+        >>> data
+            bytearray(b'hello world    \x00\x00\x00\x00\x00')
+    """
+    if not isinstance(value, str):
+        raise TypeError(f"Value value:{value} is not from Type string")
```
gijzelaerr, line 434: to be consistent, I think validating the type should be handled by a type checker like mypy, and not be done manually.

gijzelaerr (changes requested):
i don't think this affects any existing usage, so I don't object adding this to the lib. one small comment about type checking.

vmsh0:
I have made the requested change as well as fixed the reported linting issues.

Issue #380, opened 2022-08-01
In the DB class, there is no canonical way to read or write the entire DB


vmsh0:
When using the `DB` class, you provide initial data, and then you're able to provide a new underying `bytearray` to refresh the data.

While `DB_Row` allows to write to and read from the PLC to move data in and out of this bytearray (each row on its own slice), there is no way to write the entire `DB` (i.e. all the rows) at once. This is especially detrimental to our use case because we have a large number of rows, and separating the reads/writes is degrading our application performance.

As far as I can make out, when `row_offset` is 0 there is no real problem with reading and writing all the rows at once. As such, I propose the implementation of the `DB.read(client)` and `DB.write(client)` methods.

Again, I'm available & happy to submit PRs after any relevant feedback on the existing PRs and issues.

spreeker:
This is implemented.

https://www.youtube.com/watch?v=G-Gj_r2BQBk

vmsh0:
Hi,

> This is implemented. Rather Dangerous Operation and you better be sure the mapping is correct :).
> 

I absolutely agree it's a dangerous operation, just as dangerous as writing any value at all on a live system. This is why we have the `DB` class: you define your mapping and then the class sticks to it instead of you having to do that. It's an abstract model. So I argue that no, it is not implemented: currently, to perform the operation, you need to leave the abstract model and manipulate a byte buffer and some loosely-linked metadata.

Other than the purely pragmatic argument that it makes the model more expressive, it's also a matter of coherence: why is it possible to write back single rows, but not all of them? (in some cases that's _literally_ not possible, when using the `DB` abstraction, without working with non-public parts of the API or - again - manually manipulating the byte buffer and loosely-linked metadata; see #379)

> https://www.youtube.com/watch?v=G-Gj_r2BQBk

Sorry, I'm not sure what I'm supposed to be seeing in this video!

> 
> def write_data_db(dbnumber, all_data, size):

vmsh0:
Just for sake of reference, I have now documented in #379 that there is a way after all to iterate on keys, with some caveats.

Issue #379, opened 2022-08-01
In the DB class, there is no canonical way to iterate on all rows when id_field is not None


vmsh0:
When `id_field` is set to something that is not `None`, there is no canonical way or public API to iterate on all rows or even to obtain a list of the available keys.

The only way to do that currently is by accessing the `index` field, which however is not documented and supposedly can't be presumed to be part of the public API. (Furthermore, even if it was, directly accessing attributes to get data is usually bad API design, although the `@property' decorator somewhat mitigates the associated problems.)

What I propose is one or more of - with my personal preference being the first two items of the list:
- making `DB` iterable like dictionaries
- having a `DB.items()` method like that of dictionaries
- having a `DB.export()` method like that of `DB_Row`

Like for the other issues, I'm available to submit PRs. However, since I already have two PRs pending, I'm going to wait for your feedback on those and this issue before opening more.

vmsh0:
I just noticed the class has an `__iter__` method without any documented semantics. It iterates over the values. This method, in other mappable types such as `dict`, iterates over the _keys_, not the values.

Since I'm guessing it's too late to change that interface, I'm going to put forward a PR soon with the following changes:
- `keys()` and `__contains__()` methods, with `dict` semantics
- `export()` method, flattening all available data to a plain `dict`

spreeker:
Hai that might me a good idea. At the time I was in a hurry building all of this.

I do remember there where some random inconsistency between DB objects with different starting offsets. So I let go of solving that In the abstract class.

I've implemented code that every few seconds would read the entire DB and did updates to some values. The update would take 5ms. On a single value. The reading of the entire DB could take up almost a whole second and was a rather hefty operation (for a plc). Doing a key by key update is slowish.

It is as fast as possible.

My fist Implantation involved some OPC middle ware and took seconds and was unrealizable when I was reading just a few variables.

PR #385, opened 2022-08-22
DB usability improvements

vmsh0:
In reference to Issues #379 and #380, I'm opening this PR with proposed improvements for the `DB` class.

Please let me know whether there's anything I can change/improve to get this pulled.

### DIFF
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -1364,7 +1364,7 @@ class DB:
 
     Examples:
         >>> db1[0]['testbool1'] = test
-        >>> db1.write()   # puts data in plc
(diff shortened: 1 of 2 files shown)


vmsh0:
As you can see, I also added some static typing fixes that seemed appropriate. Existing code, as well as some of my new code, failed static typechecks without those fixes.

In particular, `None` values for some of the numeric parameters of `DB` where accepted for some of the instance attributes. This doesn't make sense as:
- `None` values are not semantically valid for those members
- They all have a well-defined (and useful) default

gijzelaerr (approved):

PR #386, opened 2022-08-23
Make all internal imports relative

vmsh0:
Not sure whether this is welcome, but we're using is as an internal patch and I thought it would be nice to offer it here.

This basically allows more freedom installing and moving the snap7 module around the filesystem for peculiar setups, as far as I know it's a pretty common practice for Python libraries.


vmsh0:
Not sure why some tests are failing, is it due to my changes? Will check more later.

swamper123:
Relative imports can be handy, as long as you know what you do. I made the experience, that in more nested projects, many people fall into an ImportError pit (but that is another story).

Anyway - back to the topic:  In our case I am not so happy, because of `types`. `types` is a buitin module itself, but also a module in our repo. If we want to use relative imports, a refactoring has to happen to differ both things better for readability (and maybe future conflicts) reasons. Other opinions are welcomed :)

vmsh0:
> Anyway - back to the topic: In our case I am not so happy, because of `types`. `types` is a buitin module itself, but also a module in our repo. If we want to use relative imports, a refactoring has to happen to differ both things better for readability (and maybe future conflicts) reasons. Other opinions are welcomed :)

Very good point, but as long as the built-in `types` is not used in the project I don't see it as a terrible thing. (at least compared to what it is already)

snap7/client.py
```diff
@@ -8,11 +8,11 @@
 from datetime import datetime
 from typing import List, Optional, Tuple, Union
 
-import snap7
-from snap7.common import check_error, ipv4, load_library
-from snap7.types import S7SZL, Areas, BlocksList, S7CpInfo, S7CpuInfo, S7DataItem
-from snap7.types import S7OrderCode, S7Protection, S7SZLList, TS7BlockInfo, WordLen
-from snap7.types import S7Object, buffer_size, buffer_type, cpu_statuses, param_types
+from . import types
```
gijzelaerr, line 11: sorry for the delay in review. Now we are at it, I think we should be consistent and then also import `block_types`, `S7CpuInfo` etc from `.types` and remove the `types.` prefix, OR remove all imports from `types` and append the module prefix everywhere.

gijzelaerr: i think the first option is still the 'prettiest'.

gijzelaerr (changes requested):

snap7/client.py
vmsh0: No worries for the delay. I'm would be more inclined to go with the prefix myself, but I will change the patch to go prefix-less later today.

vmsh0:
I've changed all the `types` imports to drop the prefix and just explicitly import the names.

gijzelaerr (approved):

PR #453, opened 2023-08-03
Fix #451

### LINKED ISSUE #451: Switch to pyproject.toml (pep-621)
we currently use `setup.py`, which at some point should be replaced by `pyproject.toml`.

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
https://peps.python.org/pep-0621/

nikteliy:
- Replace setup.cfg and setup.py with pyproject.toml
  - Rework workflows:
    - remove "repack wheel" scripts
    - add docker container actions
    - replace cibuildwheel actions with docker container actions + auditwheel
    - replace bdist_wheel with `build`
    - build manylinux-compatible wheels
    - move pycodestyle config to pycodestyle.yml(pycodestyle doesn't support pyproject.toml)


.github/workflows/pycodestyle.yml
```diff
@@ -19,4 +19,4 @@ jobs:
           sudo python3 -m pip install --upgrade pip
           sudo python3 -m pip install pycodestyle
       - name: Run pycodestyle
-        run: pycodestyle snap7 tests
+        run: pycodestyle snap7 tests --ignore=E501,E402,W391,W292,W503
```
gijzelaerr, line 22: shouldn't this be in the pycodestyle config?

nikteliy: It should. But do we really need to add another config file?  Instead, how about keeping this parameter as a command-line option for now and then replacing `pycodestyle` with [ruff](https://github.com/astral-sh/ruff) that supports `pyproject.toml`? It will make project's configs more centralized. Ideally, transfer all configurations to `pyproject.toml`. [These](https://github.com/gijzelaerr/python-snap7/blob/b5ff3d7676a3ae0888d11b64278a65c8432ed905/.github/workflows/flake8.yml#L24) parameters too.

gijzelaerr: whatever you prefer, i thought might have forgotten it. migrating to ruff is also fine with me.

nikteliy: I'll take care of migrating to ruff in the next PR

PR #477, opened 2023-11-07
Prepare for 1.4

gijzelaerr:
This:
 * removed `pytest-asyncio` since we don't use it (yet)
 * enables python 3.12 build
 * restructure getters and setters in utils into sub-modules
 * replace pycodestyle with ruff in Make target
 * replace pkg_resources with importlib so we don't need setuptools
 * replace the ADict with a dict since we didn't use it as a attribute dict anyway
 * improve error reporting in case of missing snap7 library
 
Todo:
finish, ruff replaced some of the getter & setter imports, and mypy is not happy with the cdata buffer we use (py 3.12)


snap7/common.py
```diff
@@ -60,7 +52,22 @@ def __init__(self, lib_location: Optional[str] = None):
                              or find_library('snap7')
                              or find_locally('snap7'))
         if not self.lib_location:
-            raise RuntimeError("can't find snap7 library. If installed, try running ldconfig")
+            import platform
```
nikteliy, line 55: `platform` has already been imported at 4th line

PR #490, opened 2024-02-24
Pythonic way property S7CpInfo and S7CpuInfo

lupaulus:
- Property for CStruct
- Internal use of create only
- little doc adjustement


snap7/types.py
```diff
@@ -254,6 +255,41 @@ class S7CpuInfo(ctypes.Structure):
         ('ModuleName', ctypes.c_char * 25)
     ]
 
+    @property
```
gijzelaerr, line 258: thank you for your contribution! But I'm not sure if this is a great idea, since it shadows the local variables which are already dynamically created. It works, but it is error-prone and might be confusing. what is the problem you are trying to solve? Are some of these fields optional for the PLC you are using?

snap7/client/__init__.py
```diff
@@ -127,7 +129,7 @@ def get_cpu_state(self) -> str:
             :obj:`ValueError`: if the cpu state is invalid.
 
         Examples:
-            >>> client.get_cpu_statE()
```
gijzelaerr, line 130: ouch, bad typo! thanks

snap7/types.py
lupaulus: Yeah, it shadows the attribute tags, it was on purpose. It was about your enhancement ticket about pythonic "classes". If we want to use the attribute you need to go to source and search for the attribute names, here there are built-in in the python object. Moreover for **S7CpuInfo**, you can direct convert the byte string values in str to facilitate the handling.

I can convert it to classic get function, if shadowing the C-struct is the problem. And add some more documentation on it ?

snap7/client/__init__.py
```diff
@@ -62,13 +62,15 @@ def __init__(self, lib_location: Optional[str] = None):
...
 
     def __del__(self):
         self.destroy()
 
-    def create(self):
-        """Creates a SNAP7 client.
+    def _create(self):
+        """
+        Creates a SNAP7 client.
+        Private Method
```
gijzelaerr, line 73: the private method comment is not required.

snap7/types.py
gijzelaerr: the problem is that this is not backwards compatible, and it might become an issue if someone wants to access the raw values without conversion. You assume all users want a string in a specific encoding, which I think might not be the case.

I'm tempted to just leave it as is and leave the decoding to the end user, unless we can come up with an elegant way to convert his automatically.

snap7/client/__init__.py
```diff
@@ -70,7 +70,6 @@ def __del__(self):
     def _create(self):
         """
         Creates a SNAP7 client.
-        Private Method
```
lupaulus: @gijzelaerr Removed

lupaulus:
WIP : Docstring formating

Instead of the Property, I will add some samples in the Docstring, because there are many devs that doesn't know how a CStruct class is working in Python. That will be more efficient for people, they will not to go to source to remember the fields.

snap7/types.py
```diff
@@ -246,6 +246,19 @@ def __str__(self) -> str:
 
 
 class S7CpuInfo(ctypes.Structure):
+    """
```
nikteliy, line 249: According to [PEP](https://peps.python.org/pep-0257/#multi-line-docstrings), a docstring should follow this format: summary, followed by a blank line, and then the rest of the content.

lupaulus: Ok, I will made it after easter,
Happy Holidays

PR #494, opened 2024-03-27
Add missing set_date function

edo2313:
As written in this [comment](https://sourceforge.net/p/snap7/discussion/general/thread/f264e6cb/#4dac), a date variable is the number of days since 1990-01-01, written as a **int**.
I can't test it right now, but it's really simple so there shouldn't be any problems...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-snap7"
-version = "1.3"
+version = "1.3.1"
 description = "Python wrapper for the snap7 library"
 authors = [
     {name = "Gijs Molenaar", email = "gijs@pythonic.nl"},
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-snap7"
-version = "1.3.1"
+version = "1.3"
 description = "Python wrapper for the snap7 library"
 authors = [
     {name = "Gijs Molenaar", email = "gijs@pythonic.nl"},
--- a/snap7/util.py
+++ b/snap7/util.py
@@ -1175,6 +1175,35 @@ def get_date(bytearray_: bytearray, byte_index: int = 0) -> date:
         raise ValueError("date_val is higher than specification allows.")
     return date_val
 
+def set_date(bytearray_: bytearray, byte_index: int, date: date) -> bytearray:
+    """Set value in bytearray to date
+
+        Notes:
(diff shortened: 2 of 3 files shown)


snap7/util.py
```diff
@@ -1175,6 +1175,35 @@ def get_date(bytearray_: bytearray, byte_index: int = 0) -> date:
...
+
+        Examples:
+            >>> data = bytearray(2)
+
+            >>> snap7.util.set_date(data, 0, date(2024, 3, 27))
+
+            >>> data
+                bytearray(b'\x30\xd8')
+        """
+    if date < date(1990, 1, 1):
```
nikteliy, line 1198: Your local variable `date` overshadow the function `date`. Implementing a small test could help avoid this oversight

snap7/util.py
```diff
@@ -1175,6 +1175,35 @@ def get_date(bytearray_: bytearray, byte_index: int = 0) -> date:
...
+        raise ValueError("date is lower than specification allows.")
+    elif date > date(2168, 12, 31):
+        raise ValueError("date is higher than specification allows.")
+    _days = (date - date(1990, 1, 1)).days
+    _bytes = struct.unpack('2B', struct.pack('>h', _days))
+    bytearray_[byte_index:byte_index + 2] = _bytes
```
nikteliy, line 1204: What if `byte_index` exceeds the length of `bytearray_`?

nikteliy (changes requested):
Thank you for your contribution. I like the detailed docstring with examples. However, it seems that the code needs some improvement; it currently doesn't work as expected. It would be great if you could also add a couple of tests to your pull request.

snap7/util.py
edo2313: This problem also exists in the other functions I looked (specifically, set_time and set_int, but seems like it's in almost all of them).
Do I need to handle that?

nikteliy: I believe not, since this is normal behavior.

nikteliy (approved):

gijzelaerr (approved):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-snap7"
-version = "1.3"
```
gijzelaerr, line 7: please don't update the version number, we will take care of that.

gijzelaerr (changes requested):

gijzelaerr (approved):

gijzelaerr:
manually merged here https://github.com/gijzelaerr/python-snap7/commit/d27e373a5aeff46a149543db5b85c6dc1af94b88

thanks!

PR #515, opened 2024-05-11
Fix read the docs & improve use of enum types

gijzelaerr:


snap7/types.py
```diff
@@ -93,55 +83,79 @@ class WordLen(Enum):
...
+
+    PE = 0
+    PA = 1
+    MK = 2
+    CT = 3
+    TM = 4
+    DB = 5
+
+
+class Block(Enum):
```
nikteliy, line 129: If you make the Block class a subclass of [IntEnum](https://docs.python.org/3/library/enum.html#enum.IntEnum), the `ctype`  property could look like this:
```python3
@property
    def ctype(self) -> c_int:
        return c_int(self)
```

tests/test_client.py
```diff
@@ -127,34 +150,34 @@ def test_read_multi_vars(self):
...
-        data_items[1].Start = ctypes.c_int32(4)
-        data_items[1].Amount = ctypes.c_int32(4)  # reading a REAL, 4 bytes
-
-        data_items[2].Area = ctypes.c_int32(S7AreaDB)
-        data_items[2].WordLen = ctypes.c_int32(WordLen.Byte.value)
-        data_items[2].Result = ctypes.c_int32(0)
-        data_items[2].DBNumber = ctypes.c_int32(db)
-        data_items[2].Start = ctypes.c_int32(8)
-        data_items[2].Amount = ctypes.c_int32(2)  # reading an INT, 2 bytes
+        data_items[0].Area = c_int32(Area.DB.value)
```
nikteliy, line 153: Here and in other places, if you use `IntEnum`, you can get rid of `.value` part
image

tests/test_client.py
```diff
@@ -51,44 +66,52 @@
...
+
+
+# noinspection PyTypeChecker,PyCallingNonCallable
 @pytest.mark.client
 class TestClient(unittest.TestCase):
     process = None
 
     @classmethod
-    def setUpClass(cls):
+    def setUpClass(cls) -> None:
```
nikteliy, line 81: It's good that you noticed the lack of type annotations. I always forget about that :)
Maybe we should make the mypy settings stricter. Making the return value annotation mandatory

tests/test_client.py
```diff
@@ -660,153 +672,150 @@ def test_check_as_completion(self, timeout=5):
...
         # Test write area with a DB
-        area = Areas.DB
+        area = Area.DB
         dbnumber = 1
         size = 1
         start = 1
         data = bytearray(b"\x11")
         wordlen, cdata = self.client._prepare_as_write_area(area, data)
-        res = self.client.as_write_area(area, dbnumber, start, size, wordlen, cdata)
+        _ = self.client.as_write_area(area, dbnumber, start, size, wordlen, cdata)
```
nikteliy, line 720: If this variable is not used, can it be removed completely?

nikteliy (commented):
Great job! But I'm not sure about changing `Areas` -> `Area`. This might break someone's code, but on the other hand it's easy to fix. So it's fine

PR #516, opened 2024-05-12
Improve typing

nikteliy:
- add protocols for snap7's functions
- add `--strict` parameter to mypy in pre-commit.conf


haronaut:
This PR will break python 3.8 usage (again).
see:
#517

gijzelaerr:
> This PR will break python 3.8 usage (again). see: #517

and again the answer is that we dropped support for <=3.8

snap7/client/__init__.py
```diff
@@ -18,11 +20,11 @@
 logger = logging.getLogger(__name__)
 
 
-def error_wrap(func):
+def error_wrap(func: Callable[..., Any]) -> Callable[..., Any]:
     """Parses a s7 error code returned the decorated function."""
 
-    def f(*args, **kw):
-        code = func(*args, **kw)
+    def f(*args: Any, **kwargs: Any) -> None:
```
gijzelaerr, line 26: isnts args always a list and kwargs always a dict?

.pre-commit-config.yaml
```diff
@@ -22,9 +22,11 @@ repos:
       - id: mypy
         additional_dependencies: [types-setuptools]
         files: ^snap7
+        args: [--strict]
```
gijzelaerr, line 25: doesn't mypy pick up settings from pyproject.toml?

https://github.com/gijzelaerr/python-snap7/blob/8612423c5729bd64ba4201703479ecd4bfbede61/pyproject.toml#L60

if so we can set strict everyone in one go.

snap7/client/__init__.py
```diff
@@ -47,10 +49,10 @@ class Client:
         >>> client.db_write(1, 0, data)
     """
 
-    _lib: Any  # since this is dynamically loaded from a DLL we don't have the type signature.
+    _lib: Snap7CliProtocol
     _read_callback = None
```
gijzelaerr, line 53: doesn't mypy complain about missign annotation here?

snap7/client/__init__.py
```diff
@@ -97,7 +101,7 @@ def destroy(self) -> Optional[int]:
         logger.info("destroying snap7 client")
         if self._lib and self._s7_client is not None:
             return self._lib.Cli_Destroy(byref(self._s7_client))
-        self._s7_client = None
+        self._s7_client = None  # type: ignore[assignment]
```
gijzelaerr, line 104: so `self._s7_client` is actually optional (can be None), you removed that above and now you need to type ignore that.

snap7/client/__init__.py
```diff
@@ -1000,7 +1004,7 @@ def wait_as_completion(self, timeout: int) -> int:
         check_error(result, context="client")
         return result
 
-    def _prepare_as_read_area(self, area: Areas, size: int) -> Tuple[WordLen, Array]:
+    def _prepare_as_read_area(self, area: Areas, size: int) -> Tuple[WordLen, "Array[_SimpleCData[int]]"]:
```
gijzelaerr, line 1007: why does this type need to be a a string?

snap7/client/__init__.py
```diff
@@ -1013,7 +1017,9 @@ def _prepare_as_read_area(self, area: Areas, size: int) -> Tuple[WordLen, Array]
         usrdata = (type_ * size)()
         return wordlen, usrdata
 
-    def as_read_area(self, area: Areas, dbnumber: int, start: int, size: int, wordlen: WordLen, pusrdata) -> int:
+    def as_read_area(
+        self, area: Areas, dbnumber: int, start: int, size: int, wordlen: WordLen, pusrdata: "Array[_SimpleCData[int]]"
```
gijzelaerr, line 1021: why does this type need to be a a string?

snap7/server/__main__.py
```diff
@@ -20,18 +21,18 @@
 logger = logging.getLogger("Snap7.Server")
 
 
-@click.command()
-@click.option("-p", "--port", default=1102, help="Port the server will listen on.")
-@click.option(
+@click.command()  # type: ignore[misc]
```
gijzelaerr, line 24: whats wrong here that it requires ignoring? maybe add a comment

snap7/server/__init__.py
```diff
@@ -177,17 +194,17 @@ def start(self, tcpport: int = 102):
...
         return self._lib.Srv_Stop(self._s7_server)
 
-    def destroy(self) -> Optional[int]:
+    def destroy(self) -> None:
         """Destroy the server."""
         logger.info("destroying server")
         if self._lib and self._s7_server is not None:
             return self._lib.Srv_Destroy(byref(self._s7_server))
-        self._s7_server = None
+        self._s7_server = None  # type: ignore[assignment]
```
gijzelaerr, line 207: also this one should be optional type

snap7/util/db.py
```diff
@@ -84,7 +86,7 @@ class DB:
     """
 
     bytearray_: Optional[bytearray] = None  # data from plc
-    specification: Optional[str] = None  # layout of db rows
+    specification: str  # layout of db rows
```
gijzelaerr, line 89: also this one should be optional i think

snap7/client/__init__.py
```diff
@@ -1239,7 +1245,7 @@ def as_tm_write(self, start: int, amount: int, data: bytearray) -> int:
         check_error(result)
         return result
 
-    def as_upload(self, block_num: int, _buffer, size) -> int:
+    def as_upload(self, block_num: int, _buffer: "_SimpleCData[Any]", size: "_SimpleCData[Any]") -> int:
```
gijzelaerr, line 1248: why do _all_ this type need to be a string?

gijzelaerr (changes requested):

.pre-commit-config.yaml
nikteliy: fixed

snap7/server/__main__.py
nikteliy: It doesn't matter anymore. I've found a package with stubs for `click`

snap7/client/__init__.py
nikteliy: Python interprets  `_SimpleCData[Any]` as indexing an element, which raises an exception . Therefore, it needs to be enclosed in quotes
Screenshot from 2024-06-02 22-42-12

snap7/client/__init__.py
nikteliy: https://github.com/gijzelaerr/python-snap7/pull/516#discussion_r1623561951

snap7/client/__init__.py
nikteliy: https://github.com/gijzelaerr/python-snap7/pull/516#discussion_r1623561951

snap7/client/__init__.py
nikteliy: If the type of `self._s7_client` is annotated as `Optional`, then `mypy` requires a check that `self._s7_client is not None` before each usage.`self._s7_client` is only None after the object is destroyed. So I decided to make the type S7Object and suppress the mypy error in just one place, instead of adding checks in many places.

image


Do we really need the `self._s7_client` to be able to be None?

snap7/server/__init__.py
nikteliy: It's the same as https://github.com/gijzelaerr/python-snap7/pull/516#discussion_r1623571102

snap7/util/db.py
nikteliy: Should it? `specification` is passed to DBRow in the constructor of the `DB` class, when it is passed to [`parse_specification`](https://github.com/gijzelaerr/python-snap7/blob/master/snap7/util/db.py#L351), which expects `str`. If you pass `None` to `parse_specification`, an exception will be raised. So I think `mypy` correctly indicates that None is not a suitable type.

image

snap7/client/__init__.py
gijzelaerr: this is just a choice, do you want to follow mypy strictly or do you trust your overrides? I don't mind some None checking.

gijzelaerr (approved):

snap7/client/__init__.py
gijzelaerr: ok lets just do this, its cleaner and it doesn't matter that much

PR #530, opened 2024-08-05
function to match the native DB TIA portal export with the python snap 7 db specification layout 

Novecento99:
Hi,

the siemens tia portal permits the export of a DB in a semi automatic method (copy and paste the view). The result is a text like this:

```
anotherTest	Byte	1.0	16#0	---	False	True	True	True	False		
test test	String	2.0	''	---	False	True	True	True	False		
lole	Byte	258.0	16#0	---	False	True	True	True	False	
```
the DB layout specification of this library is currently formatted as follow,:

```
24.0	LockAct		    BOOL
24.1	GrpErr		    BOOL
24.2	RuyToStart	    BOOL
24.3	RdyToReset	    BOOL
24.4	LocalAct	    BOOL
24.5	AutAct		    BOOL
24.6	ManAct		    BOOL
24.7	OoSAct		    BOOL
```

Since a DB can contain thousands of variables modified daily (by machine manufacturers for example) converting the native TIA export to this db layout is source of errors and fatigue

I decided to firstly create an utiliy function that convert the native tia layout to the current accepted layout as a starting point that therefore can pass the string argument as the db specification

The ideal solution could be directly change the DB layout format that snap7 library takes in and therefore the parsing block. 

let me know your thoughts

thanks

### DIFF
--- a/example/TIA_export_compatibility.py
+++ b/example/TIA_export_compatibility.py
@@ -1,0 +1,55 @@
+import re
+
+
+def fromTIAtoSnap(file_path:str):
+    """function to convert TIA export layout to Snap7 DB input layout
+    """
+    # Stringa di esempio
+    stringa =  open(file_path, 'r').read()
+
+    # Espressione regolare per isolare i primi tre valori separati da tabulazioni
+    pattern = r'^[\t]*([^\t]+)\t([^\t]+)\t([^\t]+)'
+
+    # Suddividi la stringa in righe
+    righe = stringa.strip().split('\n')
+
+    newText : str = ""
+    # Itera su ogni riga e cerca la corrispondenza
+    for riga in righe:
+        match = re.match(pattern, riga)
+        print(f"Riga:\t{riga}")
+        if match:
+            var_name, type_, index = match.groups()
+
+            print(f"var:\t{var_name.strip()}")
+            print(f"type:\t{type_.strip()}")
+            print(f"index:\t{index.strip()}")
+            newText = newText+" "+index+" "+var_name+" "+type_+"\n"
+            
+        else:
+            print("error with parse")
+            #print(f"Nessuna corrispondenza trovata nella riga: {riga}")
+        print("parsing completed")
+    return newText
+
+
+
+if __name__ == "__main__":
+    from snap7 import DB
+
+    file_path = r"example\db_layout_fromTIA.txt"
+
+    dbSpecificationLayout = fromTIAtoSnap(file_path)
+
+    print(dbSpecificationLayout)
+    db_ = DB(
+        1,
+        bytearray(1),
+        dbSpecificationLayout,
+        1,
+        1,
+
+    )
+
+    # with open("Output.txt", "w") as text_file:
+    #     text_file.write(dbSpecificationLayout)
--- a/example/db_layout_fromTIA.txt
+++ b/example/db_layout_fromTIA.txt
@@ -1,0 +1,4 @@
+	PieroAppoggiato	Bool	0.0	false	FALSE	False	True	True	True	False		
+	anotherTest	Byte	1.0	16#0	---	False	True	True	True	False		
+	test test	String	2.0	''	---	False	True	True	True	False		
+	lole	Byte	258.0	16#0	---	False	True	True	True	False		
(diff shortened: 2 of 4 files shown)


lupaulus:
Nice a have, do locally a branch the same with interpretation from Inputs/Outputs/Merk and DB with the S7 `PLCTags.xml` generate from SIMANTIC Manager and [WinCC Tag Converter](https://support.industry.siemens.com/cs/document/56078300/tagconverter-for-wincc-(tia-portal)?dti=0&lc=en-TN)

Novecento99:
> Nice ! I a have, do locally a branch, the same, with interpretation from Inputs/Outputs/Merk and DB with the S7 `PLCTags.xml` generated from SIMANTIC Manager and [WinCC Tag Converter](https://support.industry.siemens.com/cs/document/56078300/tagconverter-for-wincc-(tia-portal)?dti=0&lc=en-TN)
> 
> But with TIA, I am blocked with non-optimized DB's that aren't exported in the `PLCTags.xml`

wait let me understand better: you made a branch to ingest input/outputs/merk/db through an .xml generated by the winCCconverter. It is also able to read/write optimized dbs?

spreeker:
Hi nice work / effort.  

When I created the first version of this layout parsing code I copy pasted the DB information / the view from inspecting the DB object in siematic.manager i think.   I discovered that the layout often required manual edits amd changes. Sometimes the starting point of the entire DB object was just moved.

I did not find a reliable way to just access a DB objects layout. How DB object ended up in a plc seemed a bit unpredicatble for me.

If there is a way that can reliable work to extract that information that would be awesome!

Cheers,

Stephan.

Novecento99:
also, a discussion that I would like to brought up is about the arrays of structs. There will be multiple variables with the same name.. how to manage them ? 
A possibility, that I implemented for a specific case, is to assign a number at the end of the name which increases if there are duplicates

lupaulus:
> > Nice ! I a have, do locally a branch, the same, with interpretation from Inputs/Outputs/Merk and DB with the S7 `PLCTags.xml` generated from SIMANTIC Manager and [WinCC Tag Converter](https://support.industry.siemens.com/cs/document/56078300/tagconverter-for-wincc-(tia-portal)?dti=0&lc=en-TN)
> > But with TIA, I am blocked with non-optimized DB's that aren't exported in the `PLCTags.xml`
> 
> wait let me understand better: you made a branch to ingest input/outputs/merk/db through an .xml generated by the winCCconverter. It is also able to read/write optimized dbs?

No, optimized or non optimized db, because the export of the variable in TIA doesn't include DB's variables.
I talked with Siemens people, they say, there would be the SCADA export possibility, but it is encrypted or the DB file (.db) as plain text that need to be reinterpreted... 

For the moment, on TIA Project, I use OPC UA with the XML that can be exported from TIA or only the variable table from TIA (Inputs/Output/Merkers)

lupaulus:
> Hi, thankyou
> 
> what I want to add, is the possibility to copy and paste the view from the TIA portal to a file and therefore use it to construct a db object. I'm not familiar with the simatic manager, I've never used it actually. Does it permit to access optimized DBs?

Optimized DB's didn't exist, at the time of Simatic Manager.

lupaulus:
> also, a discussion that I would like to brought up is about the arrays of structs. There will be multiple variables with the same name.. how to manage them ? A possibility, that I implemented for a specific case, is to assign a number at the end of the name which increases if there are duplicates

In the case you want to handle custom Structs, I would prefer use OPC UA, because of the description of the Struct built in the export

Novecento99:
> > also, a discussion that I would like to brought up is about the arrays of structs. There will be multiple variables with the same name.. how to manage them ? A possibility, that I implemented for a specific case, is to assign a number at the end of the name which increases if there are duplicates
> 
> In the case you want to handle custom Structs, I would prefer use OPC UA, because of the description of the Struct built in the export

well I would prefer it too, except it seems too slow for some applications that I need to implement. For the moment, I just added "_0", "_1" , "_X" to the variabile name in case of arrays and it works

lupaulus:
> > > also, a discussion that I would like to brought up is about the arrays of structs. There will be multiple variables with the same name.. how to manage them ? A possibility, that I implemented for a specific case, is to assign a number at the end of the name which increases if there are duplicates
> > 
> > 
> > In the case you want to handle custom Structs, I would prefer use OPC UA, because of the description of the Struct built in the export
> 
> well I would prefer it too, except it seems too slow for some applications that I need to implement. For the moment, I just added "_0", "_1" , "_X" to the variabile name in case of arrays and it works

I did the same lol for arrays

Novecento99:
okay, I'm going to implement the changes then, I will submit a new cleaner PR, this one has the commits of my last one....

Issue #545, opened 2024-11-02
Implement a 100% python version, replacing the snap7 library


gijzelaerr:

lupaulus:
Nice idea

lupaulus:
@gijzelaerr do you need some help on your PR ?

gijzelaerr:
maybe later on, for me it is a tryout to use AI coding tools for auto translation from C++ to python. Once i laid the foundation to have the old python wrapper communicate with the new low level code i can use some hand porting all the functions.

hirotasoshu:
I think it's a great idea. Also, as a reference, there is a native implementation in Go: https://github.com/robinson/gos7

PR #546, opened 2024-11-02
Native python implementation

gijzelaerr:
implements #545

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,12 +16,12 @@ repos:
       - id: check-docstring-first
       - id: detect-private-key
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.10.0'
-    hooks:
-      - id: mypy
-        additional_dependencies: [types-setuptools, types-click]
-        files: ^snap7
+#  - repo: https://github.com/pre-commit/mirrors-mypy
+#    rev: 'v1.10.0'
+#    hooks:
+#      - id: mypy
+#        additional_dependencies: [types-setuptools, types-click]
+#        files: ^snap7
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: 'v0.4.2'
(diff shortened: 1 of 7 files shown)


lupaulus:
I will try to work on ISO-on-TCP Stack

gijzelaerr:
this idea is a bit in flux. I've been trying to do a full auto translation with google gemini and github co-pilot with mixed results. I have another branch somewhere that has better results, but i didn't find the time to polish it.

How do you propose to work on the iso-on-tcp stack?

lupaulus:
I propose to convert Davide's work on the basis of ISO TCP. I also have another ISO TCP integration at my disposal, which is optimised for faster communications (20ms). I used co-pilot too, but it needs cleaning and adaptation with python standards and performance (multiprocess - threadings - socket).

gijzelaerr:
thats basically what is happening in the branch, the basics should be there. What would be different? why is this alternative implementation faster?

lupaulus:
I am not sure if it faster than David exchange, but they seem to be a different usage of hardware capabilities (Processor, thread, realtime kernel)

Here [my branch](https://github.com/lupaulus/python-snap7/tree/native_python/snap7/low_level), I will continue 

I heard that Siemens will release a new CPU for the end of the year.  From what I've heard, this will remain an S7.
I'm seeking for documentation about the S7+ protocol, not sure if I can implement it.

gijzelaerr:
so my original plan was:
* reimplement c++ implementation in python, as we/you now mostly did in this branch. Aim is to make it a drop in replacement with the same/similar API, which can be a bit clunky initially.
* do a bit of testing see if it works with the emulator
* integrate it with the existing code, remove the DLL and reroute all calls that went to the DLL to the new native implementation
* Make sure test suite runs
* cleanup phase, see if we can simplify some calls
* code style and type checks passes everywhere
* done!

what do you think?

lupaulus:
Go, I work on my free time when possible. I will try to stick to this plan :)

gijzelaerr:
Btw about s7+, last time I checked it was a proprietary protocol. There might be some legal issues reverse-engineering it.but I might be wrong.

lupaulus:
Update on my progress,

I'm converting the client using Sharp7 and snap7.
Some problems with byte conversions in python, but it's fine. 
I'm not far from establishing a stable connection.

PR #569, opened 2025-09-11
Pure native python snap7 

gijzelaerr:
# Pure Python S7 Protocol Implementation

This PR replaces the C library (Snap7) dependency with a pure Python implementation of the S7 protocol.

## What's changing

We're switching to a **pure Python implementation** of the S7 protocol - no more C library dependency! This means:
- Easier installation (no compiling, no binary dependencies)
- Works on any platform Python runs on
- Easier to debug and extend

The new implementation was developed with the help of **Claude AI** (Anthropic's coding assistant), which helped implement the S7 protocol from scratch based on protocol documentation and testing.

## We need testers!

I don't own a PLC myself, so I'm looking for volunteers who can test this against real hardware.

### Compatible PLCs (S7 protocol)

If you have any of these, we'd love your help:

| Series | Models |
|--------|--------|
| **S7-1200** | CPU 1211C, 1212C, 1214C, 1215C, 1217C |
| **S7-1500** | CPU 1511, 1513, 1515, 1516, 1517, 1518 |
| **S7-300** | CPU 312, 313, 314, 315, 317, 318, 319 |
| **S7-400** | CPU 412, 414, 416, 417 |
| **LOGO!** | 0BA7, 0BA8 (with Ethernet) |
| **S7-200 Smart** | Various models |
| **WinAC** | Software PLCs |

Also compatible with S7 protocol simulators like **PLCSIM Advanced**, **NetToPLCSim**, etc.

### 1. Install the test version

```bash
pip install git+https://github.com/gijzelaerr/python-snap7.git@native
```

### 2. Configure your PLC

- Enable **"Permit access with PUT/GET"** in TIA Portal (for S7-1200/1500)
- Create two Data Blocks:
  - **DB1** "Read_only" with test values
  - **DB2** "Read_write" for write tests
- See the [test file header](https://github.com/gijzelaerr/python-snap7/blob/native/tests/test_client_e2e.py) for the exact DB structure

### 3. Run the tests

```bash
# Clone the repo
git clone -b native https://github.com/gijzelaerr/python-snap7.git
cd python-snap7

# Install with test dependencies
pip install -e ".[test]"

# Run e2e tests with your PLC settings
pytest tests/test_client_e2e.py --e2e \
    --plc-ip=YOUR_PLC_IP \
    --plc-rack=0 \
    --plc-slot=1 \
    -v --html=report.html
```

**Available options:**
| Option | Default | Description |
|--------|---------|-------------|
| `--e2e` | (required) | Enable e2e tests |
| `--plc-ip` | 10.10.10.100 | PLC IP address |
| `--plc-rack` | 0 | Rack number |
| `--plc-slot` | 1 | Slot number |
| `--plc-port` | 102 | TCP port |
| `--plc-db-read` | 1 | Read-only DB number |
| `--plc-db-write` | 2 | Read-write DB number |

### 4. Share your results

Please comment on this PR with:
- Your PLC model and firmware version
- Which tests passed/failed
- The HTML report if possible

## Quick smoke test

If you just want to do a quick test:

```python
from snap7.client import Client

client = Client()
client.connect("YOUR_PLC_IP", 0, 1)  # IP, rack, slot
print(f"Connected: {client.get_connected()}")
print(f"CPU state: {client.get_cpu_state()}")

# Read 10 bytes from DB1 at offset 0
data = client.db_read(1, 0, 10)
print(f"Data: {data.hex()}")

client.disconnect()
```

## Current status

✅ **Working:**
- Connection/disconnection
- DB read/write (all data types)
- Multi-var read/write
- CPU state, PDU length
- Order code, PLC datetime

⚠️ **May vary by PLC model:**
- SZL reads (some IDs not available on S7-1200/1500)
- Block listing
- CPU info

---

Thanks for any help! Every test against real hardware helps make this library better.


CLAUDE.md
nikteliy: Maybe this option will be more universal https://agents.md/

gijzelaerr:
it would be more useful if the comment was a bit more specific.

gijzelaerr:
This should be in a better shape now.

amorelettronico:
Hello @gijzelaerr, 

Sure, i tested the latest native branch. There are 5 failed test left. Unfortunaly I was not able to use the arguments on the commandline for the host IP and slot no, so i used the env vars instead.

Will test other hardware also if you're intersted.

[report.html](https://github.com/user-attachments/files/25471936/report.html)

amorelettronico:
Hi @gijzelaerr,

When the env var is set to use a other db then DB1,DB2 tt fails because there is a hard coded DB_READ to DB1 in:
tests/test_client_e2e.py::TestClientDBWrite::test_db_write_int
```
self = <test_client_e2e.TestClientDBWrite testMethod=test_db_write_int>

    def test_db_write_int(self) -> None:
        """Test db_write() for Int values."""
        test_value = 10
        data = bytearray(2)
        set_int(data, 0, test_value)
        self.client.db_write(DB_READ_WRITE, OFFSET_INT1, data)
    
        # Read back and verify
>       result = self.client.db_read(1, OFFSET_INT1, 2)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

amorelettronico:
I got the same results on a CPU416-2, can test on a 410-5H, 1500 and 1200. But not on a short window.

gijzelaerr:
@amorelettronico Thanks for the detailed test report! Both issues should now be fixed in 7e287d3.

### Fix 1: `test_db_get` — Protocol encoding bug

The root cause was a bug in the USER_DATA request encoding in `s7protocol.py`. The data section header for requests with non-zero payloads was using `0x0A, 0x00` instead of `0xFF, 0x09`:

- `0x0A` = "Object does not exist" (error code) → should be `0xFF` = "Data OK"
- `0x00` = null transport size → should be `0x09` = octet string

This matches the encoding used by the original Snap7 C library. Our Python test server was lenient about these values, but real PLCs (like the CPU416-2) rejected the request — and ironically the error returned (`0x0A`) was the exact value we were sending in the request header.

Fixed in `build_get_block_info_request`, `build_list_blocks_of_type_request`, `build_read_szl_request`, and `build_set_clock_request`. Functions with zero-length payloads (`build_list_blocks_request`, `build_get_clock_request`, `build_userdata_followup_request`) correctly use `0x0A, 0x00` and were left unchanged.

Additionally, `db_get()` and `db_fill()` now accept an optional `size` parameter so users can bypass `get_block_info()` on PLCs that don't support it at all.

### Fix 2: `test_db_read_bool` — Documentation clarification

Updated the DB setup documentation to explicitly state the expected byte value: `0x01` (bool0=True, bool1-7=False).

All tests and checks pass (pytest, mypy, ruff).

Could you re-test on your CPU416-2 when you get a chance?

lupaulus:
@gijzelaerr Here my output on CPU 1211C
```
metadata: {'Python': '3.12.2', 'Platform': 'Windows-11-10.0.26200-SP0', 'Packages': {'pytest': '9.0.2', 'pluggy': '1.6.0'}, 'Plugins': {'cov': '7.0.0', 'html': '4.2.0', 'metadata': '3.1.1'}}
rootdir: D:\Projekt\python-snap7
configfile: pyproject.toml
plugins: cov-7.0.0, html-4.2.0, metadata-3.1.1
collected 47 items / 46 deselected / 1 selected

tests/test_client_e2e.py::TestClientDBOperations::test_db_get
--------------------------------------------------- live log setup ----------------------------------------------------
INFO     snap7.client:client.py:109 S7Client initialized (pure Python implementation)
DEBUG    snap7.connection:connection.py:205 TCP connected to 192.168.18.235:102
DEBUG    snap7.connection:connection.py:219 Sent COTP Connection Request
DEBUG    snap7.connection:connection.py:328 Negotiated PDU size: 1024
DEBUG    snap7.connection:connection.py:330 Unsupported COTP parameter: code=0xc1, length=2
DEBUG    snap7.connection:connection.py:330 Unsupported COTP parameter: code=0xc2, length=2
DEBUG    snap7.connection:connection.py:231 Received COTP Connection Confirm
INFO     snap7.connection:connection.py:113 Connected to 192.168.18.235:102, PDU size: 1024
DEBUG    snap7.connection:connection.py:156 Sent 25 bytes
INFO     snap7.client:client.py:1855 Negotiated PDU length: 240
INFO     snap7.client:client.py:190 Connected to 192.168.18.235:102 rack 0 slot 1
---------------------------------------------------- live log call ----------------------------------------------------
DEBUG    snap7.connection:connection.py:156 Sent 37 bytes
WARNING  snap7.s7protocol:s7protocol.py:1421 USERDATA response error 0x8104: This service is not implemented on the module or a frame error was reported
SKIPPED (get_block_info not supported on this PLC: Get block info failed: Object does not exist (0x0a))
-------------------------------------------------- live log teardown --------------------------------------------------
INFO     snap7.connection:connection.py:135 Disconnected from 192.168.18.235:102
INFO     snap7.client:client.py:212 Disconnected from 192.168.18.235:102
```

Only one USERDATA Error

gijzelaerr:
Thanks for sharing the results! This looks like expected behavior - the test correctly handles the case where get_block_info (USERDATA service) isn't fully supported on the S7-1200 CPU 1211C. The 0x8104 error means "service not implemented on the module", which is typical for smaller S7-1200 PLCs that don't support all USERDATA functions.

PR #589, opened 2026-02-27
Add AsyncClient with asyncio support

gijzelaerr:
## Summary

Adds a native `AsyncClient` for async S7 communication using `asyncio` streams. This replaces the previous `asyncio.to_thread()` approach (which was not safe for concurrent use) with proper non-blocking I/O, following the pattern from [python-s7comm](https://github.com/nikteliy/python-s7comm).

### Key design decisions

- **Native async I/O**: Uses `asyncio.open_connection()` with `StreamReader`/`StreamWriter` instead of blocking sockets wrapped in threads. The event loop is never blocked.
- **`asyncio.Lock()` in `_send_receive()`**: S7 is request-response, so each send+receive cycle must be atomic. The lock serializes concurrent coroutines, making `asyncio.gather()` safe.
- **Request-embedded sequence validation**: The async client extracts the expected PDU sequence number directly from the request bytes (S7 header offset 4-5) rather than relying on the shared protocol counter, avoiding race conditions when multiple coroutines build requests concurrently.
- **`ClientMixin` for shared logic**: 14 pure-computation methods (parameter management, error text, area mapping, etc.) are shared between `Client` and `AsyncClient` via a mixin in `snap7/client_base.py`, eliminating code duplication.
- **`S7Protocol` unchanged**: The protocol layer is pure computation (struct pack/unpack) — no I/O, no async needed.

### New files

- `snap7/async_client.py` — `AsyncISOTCPConnection` (async transport) and `AsyncClient` (full-featured async S7 client)
- `snap7/client_base.py` — `ClientMixin` with shared pure-computation methods
- `tests/test_async_client.py` — 26 tests covering connection, DB read/write, area read/write, concurrent safety, multi-var operations, and synchronous helpers

### Feature parity

`AsyncClient` supports the same operations as the sync `Client`:
- `connect` / `disconnect` / `get_connected`
- `db_read` / `db_write` / `db_get`
- `read_area` / `write_area` (with chunked transfers for large payloads)
- `ab_read` / `ab_write`, `eb_read` / `eb_write`, `mb_read` / `mb_write`, `tm_read` / `tm_write`, `ct_read` / `ct_write`
- `read_multi_vars` / `write_multi_vars`
- `list_blocks` / `list_blocks_of_type` / `get_block_info`
- `get_cpu_state` / `get_cpu_info` / `get_pdu_length`
- `upload` / `full_upload` / `download`
- `read_szl` / `get_plc_datetime` / `set_plc_datetime`
- `plc_hot_start` / `plc_cold_start` / `plc_stop`
- `get_cp_info` / `get_order_code` / `get_protection`
- Parameter management, session passwords, connection params
- `async with` context manager

### Test plan

- [x] All 454 existing tests pass (428 sync + 26 new async)
- [x] Concurrent safety validated: `asyncio.gather()` with multiple reads, mixed read/write, and 10 concurrent reads
- [x] Ruff formatting checks pass

### DIFF
--- a/snap7/async_client.py
+++ b/snap7/async_client.py
@@ -7,16 +7,15 @@ in a thread pool executor via asyncio.to_thread().
 
 import asyncio
 from datetime import datetime
-from typing import Any, Callable, List, Optional, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union
 
-from ctypes import Array, c_int
+from ctypes import Array
 
 from .client import Client
 from .type import (
     Area,
     Block,
     BlocksList,
-    CDataArrayType,
     Parameter,
     S7CpInfo,
     S7CpuInfo,
@@ -24,9 +23,7 @@ from .type import (
     S7OrderCode,
     S7Protection,
     S7SZL,
-    S7SZLList,
     TS7BlockInfo,
-    WordLen,
 )
 
 


nikteliy (commented):
I’m not sure this would work without a lock.
What happens if I run two tasks without awaiting them?
asyncio.create_task(client.read_data(1, 0, 1))
asyncio.create_task(client.read_data(2, 0, 2))

gijzelaerr:
@nikteliy You were absolutely right — sorry about that, Claude hallucinated and built its own thing with `asyncio.to_thread()` instead of actually adopting your approach. I thought it had followed your native async pattern from python-s7comm, but it hadn't.

I've now properly implemented it: native `asyncio.open_connection` streams with an `asyncio.Lock()` serializing each send/receive cycle, just like you did in python-s7comm. See commit c13a241.

Thanks for catching this!

PR #711, opened 2026-04-24
Accept V1 CreateObject responses without ServerSessionVersion (GH-710)

### LINKED ISSUE #710: S7-1200 Session Setup Incomplete
Hello, I am trying to connect to a S7-1200 FW v4.2.2 TIA Portal v19.

This [table](https://python-snap7.readthedocs.io/en/latest/plc-support.html) says that my PLC uses S7CommPlus V1 and is supported.

I ran the following using the `master` branch:

```py
#!/usr/bin/env python3
from s7 import Client

if __name__ == '__main__':
  client = Client()
  client.connect("192.168.0.1", 0, 1)
  variables = client.browse()
  print(variables)
  client.disconnect()
```

Output:
```
/python-snap7-master$ ./main.py
PLC did not provide ServerSessionVersion - session setup incomplete
Traceback (most recent call last):
  File "/python-snap7-master/./main.py", line 7, in <module>
    variables = client.browse()
  File "/python-snap7-master/s7/client.py", line 302, in browse
    raise RuntimeError("browse() requires S7CommPlus connection")
RuntimeError: browse() requires S7CommPlus connection
```

Attached is a zipped pcap [S71200Filtered.zip](https://github.com/user-attachments/files/27027770/S71200Filtered.zip)

Let me know if I am doing something wrong here

gijzelaerr:
Fixes #710.

## Summary

S7-1200 F/W v4.2.2 completes the S7CommPlus handshake but **omits the `ServerSessionVersion` attribute** from its CreateObject response. We were treating any missing `ServerSessionVersion` as a session-setup failure, disconnecting the CommPlus client, and falling through to legacy — then calls that require CommPlus (`browse()`, symbolic reads, …) surface confusing errors.

Confirmed from the attached pcap: the PLC's CreateObject response (packet 16) is only 38 bytes and contains no attribute 306. That's not a parser bug — the PLC just doesn't send it.

## Why V1 is OK without it

The `ServerSessionVersion` echo is the V2+ integrity-ID handshake step. V1 connections don't track integrity IDs, so the echo is advisory. Relaxing the rule:

- V1 + no `ServerSessionVersion` → accept, `session_setup_ok=True`
- V1 + `ServerSessionVersion` present → echo it back (unchanged)
- V2 / V3 + no `ServerSessionVersion` → still fail (integrity IDs genuinely required)

Applied symmetrically in `s7/connection.py` (sync) and `s7/_s7commplus_async_client.py` (async). Decision extracted into `_finalize_session_setup` so it's unit-testable without mocking TCP.

## Test plan

- [x] 4 new unit tests in `tests/test_s7_unit.py::TestFinalizeSessionSetup`: V1-without, V2-without, V3-without, V1-with.
- [x] Full suite: 1558 passed, 88 skipped.
- [x] mypy + ruff + pre-commit clean.

@xBiggs — could you pull this branch and re-run `client.browse()` against your S7-1200 to confirm? The logged warning should disappear and `browse()` should succeed instead of falling back to legacy.

### DIFF
--- a/s7/_s7commplus_async_client.py
+++ b/s7/_s7commplus_async_client.py
@@ -163,12 +163,8 @@ class S7CommPlusAsyncClient:
 
             self._connected = True
 
-            # Step 6: Session setup - echo ServerSessionVersion back to PLC
-            if self._server_session_version is not None:
-                self._session_setup_ok = await self._setup_session()
-            else:
-                logger.warning("PLC did not provide ServerSessionVersion - session setup incomplete")
-                self._session_setup_ok = False
+            # Step 6: Session setup — echo ServerSessionVersion or accept V1 without.
+            self._session_setup_ok = await self._finalize_session_setup()
             logger.info(
                 f"Async S7CommPlus connected to {host}:{port}, "
                 f"version=V{self._protocol_version}, session={self._session_id}, "
@@ -684,6 +680,23 @@ class S7CommPlusAsyncClient:
                 offset += 1
 
         logger.debug("ServerSessionVersion not found in CreateObject response")
+
+    async def _finalize_session_setup(self) -> bool:
+        """Apply the version-specific rule for step 6 of the handshake.
+
+        - If the PLC sent ``ServerSessionVersion``, echo it back (side effect).
+        - V1 PLCs don't track integrity IDs, so the echo is optional; some
+          S7-1200 firmware (V4.2.2) omits the attribute entirely and we must
+          still accept the connection. See GH-710.
+        - V2+ strictly requires the echo; treat absence as a setup failure.
+        """
+        if self._server_session_version is not None:
+            return await self._setup_session()
+        if self._protocol_version == ProtocolVersion.V1:
+            logger.debug("V1: no ServerSessionVersion from PLC (accepted)")
+            return True
+        logger.warning("PLC did not provide ServerSessionVersion - session setup incomplete")
+        return False
 
     async def _setup_session(self) -> bool:
         """Echo ServerSessionVersion back to the PLC via SetMultiVariables."""
--- a/s7/connection.py
+++ b/s7/connection.py
@@ -201,12 +201,8 @@ class S7CommPlusConnection:
             # CreateObject always uses V1 framing
             self._create_session()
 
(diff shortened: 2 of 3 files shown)


gijzelaerr:
Holding this. @xBiggs confirmed on #710 that even with V1 session setup fixed, every subsequent request fails with `ServiceUnknownRequest` — V1 S7-1200 firmware doesn't implement the V2+ opcodes (`EXPLORE`, `GET_MULTI_VARIABLES`, `SET_MULTI_VARIABLES`) that our client uses unconditionally.

Merging this alone would actually regress V1 users under AUTO mode: currently the handshake fails and they fall through to legacy (which works if PUT/GET is enabled); with this merged, they'd succeed the handshake and then hit a wall on every call.

Planning to expand this PR to also decline S7CommPlus when V1 is negotiated, so AUTO falls back to legacy transparently — then a proper V1 request path in a separate PR. Waiting on a pcap from @xBiggs before pushing the expansion.

See [this comment](https://github.com/gijzelaerr/python-snap7/issues/710#issuecomment-4314681811) for the full diagnosis.

gijzelaerr:
Closing — the premise of this PR is wrong. After analysing @xBiggs's captures (see [#710 comment](https://github.com/gijzelaerr/python-snap7/issues/710#issuecomment-4314681811-FOLLOWUP)), V1 PLCs *do* require the `ServerSessionVersion` echo to unlock data operations; when the PLC doesn't send the attribute, the session is genuinely incomplete and every subsequent request fails with `ERROR2 (0x05A9)`. That's exactly what `s7/connection.py:850-852` documents:

> Without this step, the PLC rejects all subsequent data operations with ERROR2 (0x05A9).

The original behaviour — warn, return `session_setup_ok=False`, fall back to legacy — was correct. This PR turns the guard off and forces the unified client to stay in CommPlus mode where every call then fails. That's a regression for V1 PLC users, not a fix.

What's actually needed to make `browse()` and friends work on V1-initial S7-1200s is the V2 legitimation handshake (visible in the TIA Portal capture: V2 `SET_MULTI_VARIABLES` with crypto material right after `CreateObject`). That's its own substantial piece of work and will be tracked separately.

Issue #735, opened 2026-06-16
S7-1200 Q/I/M symbolic reads: findings from real hardware testing (FW V4.5)


tommasofaedo:
# S7-1200 Q/I/M symbolic reads: findings from real hardware testing (FW V4.5)

## Hardware / software setup

- **PLC**: Siemens S7-1200 CPU 1214C DC/DC/DC, firmware V4.5
- **Module**: DI8/DQ6 (digital inputs 0–7, digital outputs Q0.0–Q0.5)
- **Config**: no password, no Put/Get, port 102
- **Library**: python-snap7 3.0.0 (`s7` S7CommPlus package)
- **Reference capture**: Wireshark capture of EXOR HMI communicating via AGLink
  (confirms wire-level S7CommPlus V1 protocol details)

We tested `read_symbolic` / `write_symbolic` against a live PLC and compared
results with captured EXOR HMI traffic. Below are our findings. Some bugs were
already fixed locally; one unfixed bug remains. All findings are from empirical
testing, not reverse-engineering documentation.

---

## Bug 1 — `_build_symbolic_write_payload` uses wrong `sub_area` for I/Q/M writes

**Status: still present in 3.0.0 — needs a fix**

`_build_symbolic_read_payload` correctly uses `sub_area=3736` (0xE98) for
non-DB areas (I/Q/M). But the write counterpart uses `CONTROLLER_AREA_VALUE_ACTUAL`
(2551), which causes the PLC to reject the write with error `0x4D0013`.

```python
# _build_symbolic_write_payload — CURRENT (broken for I/Q/M)
if access_area >= 0x8A0E0000:
    access_sub_area = Ids.DB_VALUE_ACTUAL
else:
    access_sub_area = Ids.CONTROLLER_AREA_VALUE_ACTUAL  # wrong: 2551

# SHOULD BE (matching read side)
if access_area >= 0x8A0E0000:
    access_sub_area = Ids.DB_VALUE_ACTUAL
else:
    access_sub_area = 3736  # 0xE98 — confirmed from EXOR AGLink capture
```

The value `3736` was confirmed by byte-comparing captured EXOR HMI
GetMultiVariables / SetMultiVariables frames with the library's output.

---

## Bug 2 — `_parse_read_response` misreads single-byte scalar as PLC error

**Status: fixed locally — sharing for upstream**

For a single BOOL or USINT variable, the PLC returns an 8-byte response:

```
[...]

PR #740, opened 2026-06-17
fix: handle single-byte scalar in _parse_read_response

### LINKED ISSUE #737: fix: _parse_read_response misreads single-byte scalar as PLC error
## From #735 — Bug 2

For single BOOL or USINT variables, the PLC returns an 8-byte response:

```
[value] [0x00] [0x00] [0x04] [0x00] [0x00] [0x00] [0x00]
```

The parser reads the first VLQ as `return_value`. When `value=1` (BOOL TRUE), VLQ decodes as `1` (non-zero) → parser logs `"PLC returned error: 1"` and returns `[]`. Every BOOL TRUE read raises `RuntimeError("Symbolic read failed")`.

### Fix

Special-case the single-byte scalar response format: when the response ends with the fixed 6-byte suffix `00 04 00 00 00 00` and the body before the suffix is exactly 2 bytes (value + `0x00`), return the first byte as the scalar value.

### Hardware tested on
- S7-1200 CPU 1214C DC/DC/DC, FW V4.5

### Related
- #735 (parent issue)

gijzelaerr:
## Summary

- Fix `_parse_read_response` misreading single-byte BOOL/USINT values as PLC errors on S7-1200
- S7-1200 wraps bare scalars in `[value 0x00 | 00 04 00 00 00 00]`; the VLQ parser decoded `value=1` (BOOL TRUE) as a non-zero return code, causing every TRUE read to raise `RuntimeError`
- Add early detection for this 8-byte format before the VLQ error-code path

Fixes #737 (from #735 Bug 2)

## Test plan

- [x] Added `test_single_byte_scalar_bool_true`, `test_single_byte_scalar_bool_false`, `test_single_byte_scalar_usint` to `TestParseReadResponse`
- [x] All existing tests pass (`pytest tests/test_s7_unit.py`)
- [x] mypy, ruff, pre-commit all clean
- [ ] Needs real hardware validation on S7-1200 FW V4.5

### DIFF
--- a/s7/_s7commplus_client.py
+++ b/s7/_s7commplus_client.py
@@ -455,6 +455,10 @@ class S7CommPlusClient:
 
 # -- Request/response builders (module-level for reuse by async client) --
 
+# S7-1200 wraps a single BOOL/USINT in [value 0x00 | 00 04 00 00 00 00].
+# The leading byte is misread by VLQ as a non-zero return code.
+_SCALAR_RESPONSE_SUFFIX = bytes.fromhex("000400000000")
+
 
 def _build_read_payload(items: list[tuple[int, int, int]]) -> bytes:
     """Build a GetMultiVariables request payload.
@@ -499,6 +503,12 @@ def _parse_read_response(response: bytes) -> list[Optional[bytes]]:
         List of raw bytes per item (None for errored items)
     """
     offset = 0
+
+    # S7-1200 single-byte scalar: [value 0x00 | 00 04 00 00 00 00]
+    if response.endswith(_SCALAR_RESPONSE_SUFFIX):
+        body = response[: -len(_SCALAR_RESPONSE_SUFFIX)]
+        if len(body) == 2 and body[1] == 0x00:
+            return [bytes(body[:1])]
 
     return_value, consumed = decode_uint64_vlq(response, offset)
     offset += consumed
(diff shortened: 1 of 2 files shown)


tommasofaedo:
---

```
Tested on real hardware: Siemens S7-1200 CPU 1212C DC/DC/DC, firmware V4.5.

Without this fix, reading a BOOL tag via GetMultiVariables returns a response
like `01 00 00 04 00 00 00 00`. The VLQ parser reads the leading `01` as a
non-zero return code and raises an error, even though the variable value is
simply TRUE (0x01).

With this fix applied, the following reads all work correctly:

- BOOL tag = TRUE → returns `b'\x01'` ✓  
- BOOL tag = FALSE → returns `b'\x00'` ✓  
- BYTE / USINT tag → returns the correct 1-byte value ✓  

I confirmed this by reading Q0.0 (BOOL) and QB0 (BYTE) from the Q area via
GetMultiVariables with sub_area=3736. Both values matched the physical LED
state on the PLC's output module.

The unit tests added in this PR match the exact wire format I observed in
captures from this PLC. Fix is correct and working on FW V4.5.
```

---

gijzelaerr (commented):
## Review

Clean fix. The early detection correctly identifies the S7-1200 single-byte scalar format `[value 0x00 | 00 04 00 00 00 00]` before the VLQ parser misinterprets the leading byte as a non-zero error return code.

The `len(body) == 2` guard prevents false positives on multi-item responses that happen to end with the same suffix. Tests cover BOOL TRUE, BOOL FALSE, and USINT — the three key cases.

No security concerns, no architectural impact, S7CommPlus only.

**Ready to merge.**

PR #767, opened 2026-07-10
refactor(examples): parameterize browse_tags.py, drop unittest.mock hack

tommasofaedo:
Follow-up cleanup for `examples/browse_tags.py` (merged in #756).

## Why

In your #756 review you flagged two blockers — the hardcoded PLC IP and the `unittest.mock` hack — and later said you'd pushed fixes for them. That cleanup commit didn't make it into the merge (#756 landed with `commits: 1`), so `examples/browse_tags.py` in `master` still has both. This PR applies them for real, and folds in a small correctness fix.

## Changes

- **argparse CLI** — positional `host`, optional area letters, `--port`, `--tls`, `--password`. No more module-level `PLC_HOST = "192.168.5.11"`.
  ```
  python browse_tags.py 192.168.5.11            # all areas, plaintext
  python browse_tags.py 192.168.5.11 --tls      # TLS-only PLCs (FW V4.1 "G2")
  python browse_tags.py 10.0.0.5 --tls I Q      # subset of areas
  ```
- **Remove `unittest.mock`** — the old code wrapped `connect()` in `mock.patch.object(S7CommPlusConnection, "_post_auth_legitimation", ...)`. `connect()` already wraps the optional post-auth legitimation in a `try/except` that only logs a warning when it's skipped (no-password PLCs, or the plaintext V3 path where legitimation needs TLS), so the mock was redundant. Verified the browse works unchanged without it.
- **I-area address fix** — the `43` in the FDICT sample `%I43.x` is not the real tag address (only the bit digit is a literal); rebuild `%I{ByteOffset}.{bit}` so `input_1` resolves to `%I0.0` instead of `%I43.0`.
- **PoC disclaimer** in the module docstring (FDICT is firmware-family specific).

## Tested on

Re-validated on a **S7-1200 CPU 1212C, FW V4.1 ("G2")**, over TLS 1.3 (mandatory on that unit — the FW V4.5 unit from #756 is no longer available to me, see #753):

```
$ python browse_tags.py <host> --tls
>> Area I  (RID=80) ... input_1   Bool  %I0.0  0  10
>> Area Q  (RID=81) ... output_1  Bool  %Q0.1  0   9
>> Area M  (RID=82) ... merker_1  Bool  %M0.2  0   9
```

All three project tags resolve to their exact TIA Portal addresses.

`ruff format` / `ruff check` clean. Example-only — no library code touched.

### DIFF
--- a/examples/browse_tags.py
+++ b/examples/browse_tags.py
@@ -1,54 +1,48 @@
-"""
-browse_tags.py — Read symbolic I/Q/M tags from an S7-1200 FW V4.5 PLC.
+"""browse_tags.py — Read symbolic I/Q/M tags from an S7-1200 PLC (V3 protocol).
 
-Technique: EXPLORE + decompression with a partial preset dictionary
-(Adler-32 0xce9b821b, 594 of 32768 bytes reconstructed via oracle analysis).
-The same FDICT is used for all three areas (I, Q, M) — confirmed on
-independent Wireshark pcapng captures.
+PROOF OF CONCEPT — not a supported library feature.
 
-Results on S7-1200 CPU 1212C DC/DC/DC, FW V4.5 (40-tag project):
+Technique: EXPLORE + zlib decompression with a partial preset dictionary
+(FDICT, Adler-32 0xce9b821b) reconstructed via "oracle" analysis. The same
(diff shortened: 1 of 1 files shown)


gijzelaerr (approved):
thanks! sorry merged that too quickly, you right

Issue #769, opened 2026-07-11
Split S7CommPlus into standalone s7commplus package, remove unified s7 client


gijzelaerr:
## Summary

The current `s7/` package tries to unify legacy S7 (`snap7`) and S7CommPlus behind a single `Client` that auto-negotiates the protocol. In practice the two protocols differ too much — different capabilities (browse is S7CommPlus-only), different connection requirements (TLS, SessionKey handshake), and different failure modes. The silent fallback from S7CommPlus to legacy creates confusing errors (#760, #710) and makes it hard for users to understand what's happening.

## Plan

1. **Create a new top-level `s7commplus/` package** containing the S7CommPlus-specific code currently in `s7/`:
   - `_s7commplus_client.py` → `s7commplus/client.py`
   - `_s7commplus_async_client.py` → `s7commplus/async_client.py`
   - `_s7commplus_server.py` → `s7commplus/server.py`
   - `s7/connection.py` → `s7commplus/connection.py`
   - `s7/protocol.py` → `s7commplus/protocol.py`
   - `s7/codec.py`, `s7/vlq.py`, `s7/legitimation.py` → `s7commplus/`

2. **Keep `snap7/` unchanged** — it remains the legacy S7 package for S7-300/400 and PUT/GET on S7-1200/1500.

3. **Shared code** (TPKT/COTP transport in `snap7/connection.py`) stays in `snap7` — `s7commplus` imports from it.

4. **Remove the unified `s7/` package** — no more `s7/client.py` with auto-negotiation and silent fallback. Users explicitly choose `snap7.Client` or `s7commplus.Client`.

5. **Update examples, docs, tests, and CLAUDE.md** to reflect the new structure.

## Motivation

- Users should explicitly choose their protocol rather than getting silent fallback behavior
- The two protocols have fundamentally different capabilities and requirements
- Cleaner API surface — no more "browse() requires S7CommPlus connection" errors after silent fallback
- Easier to document and maintain each protocol independently

## Timing

Do this after all current open PRs are merged (#761, #763, #764, #768) since they touch `s7/client.py` and `s7/connection.py` heavily.

## Breaking change

This removes the `s7` package entirely. Users importing from `s7` will need to switch to either `snap7` or `s7commplus`. This should be part of a major or minor version bump with migration notes.

Issue #718, opened 2026-05-01
Verify S7CommPlus V2/V3-with-TLS against a real PLC + add HMAC integrity if needed


gijzelaerr:
## What's already on master

The unified ``Client`` already implements V2/V3 with TLS, end-to-end against the emulated server (``s7/_s7commplus_server.py``). 46 tests in ``test_s7_v2.py`` and ``test_s7_tls.py`` cover the connection path, IntegrityId tracking, and password legitimation. ``connect(use_tls=True, tls_cert=..., tls_key=..., tls_ca=...)`` works.

PR #719 closes the documentation gap.

## What's actually still open (and needs hardware)

1. **Verify against a real PLC.** The emulated server is one half; a TIA Portal V17+ S7-1500 (or FW ≥ 4.5 S7-1200) is the other. Without that, we can't tell whether the existing implementation produces wire-correct frames.

2. **HMAC-SHA256 fragment integrity (if the real PLC requires it).** The current code emits an incrementing ``IntegrityId`` counter, not an HMAC. Per the Rogue7 paper (Bitan/Biham et al., Black Hat USA 2019) the V3 protocol expects each fragment to carry an ``HMAC-SHA256`` of the fragment under a session key derived from the TLS exporter secret. The emulated server doesn't enforce this. If a real PLC rejects our connection past the handshake, this is the most likely missing piece.

3. **A repro pcap.** Capture TIA Portal V17+ doing a basic ``db_read`` against a target PLC, save as a ``.pcapng`` test fixture, write a small comparison test that diffs our handshake bytes against TIA's frame-by-frame. Mostly a diagnostic, but would let us verify the implementation without needing live hardware on every CI run.

Each of these can be a separate small issue when someone has the hardware. Until then, this stays open as the umbrella.

## Out of scope here

- HarpoS7-style SessionKey handshake for V1-initial S7-1200 (FW < 4.5) — that's #710 / #717, archived under ``research/harpo-port-incomplete``.

gijzelaerr:
@ale-rinaldi — you've done most of the V2/V3 TLS work already (PRs #745, #752). Do you have a setup where you could verify V2 or V3-with-TLS end-to-end and check whether HMAC integrity is needed? Even a quick "it works" / "it doesn't" on your S7-1500 would help us close this one out.

ale-rinaldi:
I finally got to test this against real hardware. I have two Siemens PLCs on hand:

[...]

PR #774, opened 2026-07-16
refactor: split s7/ into standalone s7commplus/ package

gijzelaerr:
## Summary

Implements #769 — splits the S7CommPlus protocol implementation out of the unified `s7/` package into a standalone `s7commplus/` package and deletes the auto-negotiating unified wrappers.

- **`s7commplus/` package created** with the S7CommPlus implementation renamed from internal modules to clean public names
- **`s7/` package repurposed** as an alias for `snap7` — `from s7 import Client` is now the recommended way to use the legacy S7 protocol (replacing `from snap7 import Client`, which still works)
- Users now explicitly choose `s7.Client` (legacy S7-300/400) or `s7commplus.Client` (S7CommPlus S7-1200/1500)

## File moves

| Old path | New path |
|----------|----------|
| `s7/_s7commplus_client.py` | `s7commplus/client.py` |
| `s7/_s7commplus_async_client.py` | `s7commplus/async_client.py` |
| `s7/_s7commplus_server.py` | `s7commplus/server.py` |
| `s7/{connection,protocol,codec,vlq,legitimation,typeinfo}.py` | `s7commplus/` |

## Deleted (unified wrappers)

- `s7/client.py` — auto-negotiating Client with `Protocol.AUTO`
- `s7/async_client.py`, `s7/server.py` — unified async client and dual-protocol server
- `s7/_protocol.py` — `Protocol` enum (`AUTO`/`LEGACY`/`S7COMMPLUS`)
- `s7/partner.py`, `s7/util.py` — pure re-exports of snap7 modules
- `tests/test_s7_unified.py` — tested the deleted unified wrappers

## Updated

- **10 test files** — all `from s7.` → `from s7commplus.`
- **4 example files** — renamed and rewritten for the new API
- **14 .rst doc files** — updated to recommend `s7` over `snap7` for legacy protocol
- **Config**: pyproject.toml, tox.ini, Makefile, CLAUDE.md, README.rst, CHANGES.md

## Breaking changes (vs unreleased 4.0-dev only — 3.x is unaffected)

- `from s7 import Client` now returns the legacy `snap7.Client` instead of the auto-negotiating S7CommPlus client — use `from s7commplus import Client` for S7CommPlus
- `Protocol` enum removed — no more `Protocol.AUTO`/`Protocol.S7COMMPLUS`
- Unified `s7.Server` removed — use `s7commplus.Server` or `s7.server.Server` directly

## Test plan

- [x] 1558 tests pass, 82 skipped
- [x] ruff check + format clean
- [x] mypy clean (pre-existing errors only)
- [x] pre-commit all hooks pass


gijzelaerr:
@ale-rinaldi @Erik1000 @EliTheCoder @xBiggs — this PR splits the unified `s7/` package into a standalone `s7commplus/` package per #769. The auto-negotiating client that caused confusing fallback errors (#760, #710) is gone. Users now explicitly choose:

- `from snap7 import Client` — legacy S7 (S7-300/400, PUT/GET)
- `from s7commplus import Client` — S7CommPlus (S7-1200/1500)

There's also a new `s7` alias that re-exports `snap7` (so `from s7 import Client` still works for the legacy protocol — first step toward dropping the `snap7` name).

Would appreciate your review on whether the new package structure makes sense and nothing got lost in the move. The PR description has the full file mapping.

Erik1000:
Guess it makes sense if the protocols differ so much and its definitely better than catching exceptions everywhere.

ale-rinaldi:
Overall I'm in favour of the direction — a few thoughts, since you asked.

**The split itself:** 👍 Explicit `s7commplus.Client` vs `snap7.Client` and dropping the auto-negotiating wrapper is the right call. The silent fallback to legacy PUT/GET was a genuine footgun (Erik's #760, and #710) — good to see it gone.

**The HarpoS7 `session_auth` import:** this is really the bigger half of the PR, and I'd treat it separately from the restructure:

- **License is fine** — HarpoS7 is MIT, same as here. But MIT still requires carrying the upstream copyright/notice; the transpiled files cite the source path but I'd add HarpoS7's LICENSE/copyright under `session_auth/` (or a NOTICE) to be clean.
- **Two changes in one PR.** A pure package move is reviewable at a glance; ~97k lines of auto-generated Family-0 monoliths are not. Splitting "restructure" and "session_auth import" into two PRs would make each reviewable — and let the crypto import get the focused scrutiny it deserves.
- **Reproducible verification.** The files say "vector-verified against HarpoS7 test blobs" — are those blobs + `tools/transpile_harpo_monolith.py` wired into CI so the transpilation is re-verifiable? Otherwise the guarantee only held on the machine that generated them.
- **Wiring.** `legitimation.py` doesn't reference `session_auth`; it'd help to note where Family-0 is actually used (V2 integrity? session-key derivation?) and whether all 47 modules are needed, so no dead weight ships.

**One inconsistency worth a look — the `s7` → `snap7` alias.** The PR's stated goal is "no more silent fallback to legacy," but aliasing `s7` to `snap7` means `from s7 import Client` keeps working while now returning the *legacy* client instead of the S7CommPlus one it used to — a silent protocol switch for exactly the existing `s7` users, plus `s7.Protocol` disappears. That's the same class of surprise the PR is trying to remove. A clean `ImportError`/deprecation from `s7` pointing at `s7commplus`/`snap7` might be more consistent than silently handing back a different-protocol client. (For context, our probe reader is one of those `s7.Client()` + `Protocol.S7COMMPLUS` users — we'll migrate to `s7commplus`, no problem.)

**Offer:** the session-auth path is exactly what's still failing for us on real hardware (the legitimation challenge fetch in #718). We have an S7-1200 (1211C, FW4.6) and an S7-1500 (1507S, FW21.9) on hand — happy to run this branch against both and report whether legitimation/integrity actually completes end-to-end, which the emulated-server tests can't cover. Just say the word.

gijzelaerr:
ah wait, i didn't realise the session auth stuff was also in this PR. going to take that out, it is not ready yet and quite a mess.

gijzelaerr:
Thanks both for the feedback.

@Erik1000 — agreed, explicit protocol choice is much better than catching surprises.

@ale-rinaldi — good review, addressing each point:

**session_auth:** Removed from this PR entirely — it was accidentally included via the base branch. It'll come as a separate PR with proper license attribution, CI-wired transpilation verification, and documentation on where Family-0 plugs in. Your notes on all of those are spot-on and will be addressed there.

**The `s7` alias concern:** Fair point, but since `s7` was never in a released version (only existed in unreleased 4.0-dev), there are no real-world `s7.Client()` + `Protocol.S7COMMPLUS` users to surprise. We've decided to make `s7` the recommended name for the legacy protocol going forward — `from s7 import Client` is now the primary recommendation in all docs, with `snap7` kept as an undeprecated backwards-compatible alias. So the naming is now:

- `from s7 import Client` — legacy S7 (recommended)
- `from snap7 import Client` — same thing, still works
- `from s7commplus import Client` — S7CommPlus

No cross-wiring between `s7` and `s7commplus`.

**Hardware testing offer:** Would love that — let's pick it up on the session_auth PR once it's ready. Having real S7-1200 FW4.6 + S7-1500 FW21.9 coverage would be invaluable for the legitimation path.

Issue #757, opened 2026-06-25
S7-1200 FW V4.5: I/Q/M symbolic tag browse via oracle-reconstructed preset dictionary (33/40 tags complete)


tommasofaedo:
```
## Problem

On S7-1200 firmware V4.5 (V3 protocol), EXPLORE requests for the I/Q/M
areas (RIDs 80/81/82) return a zlib blob protected by a Siemens preset
dictionary (magic `78 7D`, FDICT flag set, dict Adler-32 `0xce9b821b`).
Python's `zlib.decompress()` returns `Z_NEED_DICT` — the dictionary is
embedded in TIA Portal and not published by Siemens.

As a result, symbolic tag names, data types, logical addresses, and byte
offsets are unavailable via `browse()` for I/Q/M areas on V3 PLCs
(reported as a known limitation in PR #742 / the browse PR).

## Solution — oracle technique

We reconstructed 594 of 32768 FDICT bytes using an "oracle" approach:
inflate the same blob four times with four synthetic test dictionaries
(all-zeros, all-0xFF, `i%256`, `i>>8`). A byte that is identical in all
four outputs is a literal; a byte that differs reveals the FDICT position
it was copied from: `position = (B_output << 8) | A_output`.

The **same FDICT** (Adler-32 `0xce9b821b`) is used for all three areas
(I, Q, M) — confirmed on three independent Wireshark pcapng captures.

With 594 FDICT positions known, `_extract_tags()` anchors on always-literal
ID values and recovers Name/DataType/LogicalAddress/ByteOffset from a
context window before each ID.

### Byte-type fallback (I/Q areas)

LogicalAddress reconstruction by exhaustion:
- **Bool** tags → FDICT encodes `LogicalAddress="%I43.{bit}"` (garbled area
  letter, correct bit); reconstruct as `%{area}{ByteOffset}.{bit}`.
- **Word/Int** tags → `%IW` / `%QW` are literal in the blob; append
  ByteOffset to get `%IW{N}` / `%QW{N}`.
- **Byte** tags → only remaining type; oracle confirms LogicalAddress value
  is not encoded. Reconstruct as `%IB{ByteOffset}` / `%QB{ByteOffset}`.

### Structural limit — M area (confirmed by pcapng oracle)

Oracle analysis of Wireshark captures of all 15 M area tags shows the
deflate stream uses an **identical sequence** for Bool, Byte, and Word
addresses. It is not possible to distinguish `%MB` from `%MW` from the
[...]

PR #776, opened 2026-07-16
feat: zlib preset dictionary decompressor for S7CommPlus blobs

gijzelaerr:
## Summary

- Adds the 19 known Siemens zlib preset dictionaries (24KB total) extracted from [thomas-v2/S7CommPlusDriver](https://github.com/thomas-v2/S7CommPlusDriver) (LGPL-3.0)
- Adds `blob_decompressor.py` with `decompress_blob()` that auto-selects the correct dictionary by Adler-32 checksum
- Replaces the partial oracle-based dictionary reconstruction from #757 with complete, verified dictionaries

The key dictionary for #757 is `IntfDescTag` (Adler-32 `0xce9b821b`, 1178 bytes) — used for I/Q/M area tag definitions on V3 PLCs (S7-1200 FW V4.5+). With the complete dictionary, decompression produces clean XML that can be parsed with standard tools instead of regex extraction from garbled output.

Addresses #757.

## Test plan

- [ ] Verify all 19 dictionary Adler-32 checksums match
- [ ] Decompress standard (no-dict) and preset-dict blobs
- [ ] Unknown dictionary raises clear error
- [ ] `find_and_decompress()` scans raw payloads for zlib streams
- [ ] @tommasofaedo: test against S7-1200 FW V4.5 I/Q/M EXPLORE responses


tommasofaedo:
Independently validated `find_and_decompress()` from this branch against a **live S7-1200 G2 (FW V4.1)** — it works out of the box. 🎉

I no longer have the G1 FW V4.5 unit from #757, so **all testing here is on the G2 only** — but it turns out the G2 declares the exact same preset dictionary. All three EXPLORE areas carry a `CMF=0x78 FLG=0x7d` stream with dict Adler-32 `0xce9b821b` (`IntfDescTag`), and `find_and_decompress(full_explore_payload)` returns clean, zero-null XML:

```xml
<?xml version="1.0" encoding="utf-8"?><IdentContainer>
<Ident Name="mtag_word" Scope="Global" LID="13"><SimpleType>Word</SimpleType><Access SubClass="SimpleAccess"><SimpleAccess ByteNumber="102" Width="Word" Range="Memory" /></Access></Ident>
...
</IdentContainer>
```

### Follow-up wiring (the part you flagged as TODO)

Since the XML is clean, the tag layer is tiny. Mapping `Range`+`Width`+`ByteNumber`+`BitNumber` to TIA syntax:

```python
import xml.etree.ElementTree as ET
from s7commplus.blob_decompressor import find_and_decompress

_RANGE = {"Input": "I", "Output": "Q", "Memory": "M"}
_WIDTH = {"Byte": "B", "Word": "W", "DWord": "D", "LWord": "L"}

def _address(rng, width, byte_no, bit_no):
    letter = _RANGE.get(rng, "?")
    if width == "Bit":
        return f"%{letter}{byte_no}.{bit_no or 0}"
    return f"%{letter}{_WIDTH.get(width, '')}{byte_no}"

def parse_ident_container(xml_text):
    tags = []
    for ident in ET.fromstring(xml_text).findall("Ident"):
        acc = ident.find("./Access/SimpleAccess")
        if acc is None:
            continue
        byte_no = int(acc.get("ByteNumber", "0"))
        bit_raw = acc.get("BitNumber")
        tags.append({
            "name": ident.get("Name"),
            "type": ident.findtext("SimpleType"),
            "lid": int(ident.get("LID", "0")),
            "address": _address(acc.get("Range"), acc.get("Width"),
                                byte_no, int(bit_raw) if bit_raw else None),
        })
    return tags
```

Result against the running G2 (verified against TIA Portal):

| Name | Type | Address | LID |
|------|------|---------|-----|
| input_1 | Bool | %I0.0 | 10 |
| output_1 | Bool | %Q0.1 | 9 |
| merker_1 | Bool | %M0.2 | 9 |
| mtag_byte | Byte | %MB100 | 11 |
| mtag_word | Word | %MW102 | 13 |
| mtag_dword | DWord | %MD104 | 14 |
| mtag_bool | Bool | %M108.0 | 16 |

This makes the whole oracle pipeline from #757 (`_build_fdict` / regex `_extract_tags` / symbolic-READ width workaround) obsolete — the `<SimpleType>` now carries the real M widths directly.

I'm opening a small stacked PR on top of this branch that adds a non-invasive `tag_browser.py` (the parser above + an EXPLORE-per-area helper), so you can wire it into `browse()` however you prefer — or just fold it into this PR. Whatever's least noisy for you. Only G2-validated on my end, flagging that honestly. Thanks again for the complete dictionaries.

PR #780, opened 2026-07-20
feat: tag_browser — parse I/Q/M tags and DB/FB/OB interfaces from decompressed EXPLORE XML

tommasofaedo:
## Summary

Follow-up to #776. That PR added the preset-dictionary decompressor; this one adds the small, **non-invasive** layer on top that turns a decompressed EXPLORE response into structured records — for I/Q/M symbols **and** DB / FB / OB / UDT block interfaces. This is the wiring flagged as TODO in #776 / #757.

All of it lives in `s7commplus/tag_browser.py` and `tests/test_tag_browser.py` and touches no existing files, so you can merge it as-is, call the helpers from `browse()` yourself, or fold it into #776 — whatever is least noisy.

### I/Q/M symbolic tags
- `parse_ident_container()` / `tags_from_explore()` — decompress the `<IdentContainer>` blob (`IntfDescTag` dict, `0xce9b821b`) and map `Range`+`Width`+`ByteNumber`+`BitNumber` to TIA syntax (`%I0.0`, `%Q0.1`, `%MB100`, `%MW102`, `%MD104`) as `Tag` records.

### DB / FB / OB / UDT interfaces
- `parse_block_interface()` / `block_interface_from_explore()` — decompress the `<BlockInterface>` blob (`DebugInfo_IntfDesc` dict, `0x66052b13`) into `Member` records (name, type, LID, offset, section). A block EXPLORE bundles several preset-dict streams, so it targets the interface dict explicitly rather than taking the first zlib stream.
- **Types.** A `<Member>`'s type is a RID reference `0x0200_00XX` where `XX` is a Siemens SoftDataType id (`SOFTDATATYPE` map transcribed from thomas-v2/S7CommPlusDriver `Core/Softdatatype.cs`, the same source as the shipped dictionaries). Complex members (arrays, UDTs, strings) already carry a ready-made, human-readable `Type` attribute (`Array[0..5] of IEC_TIMER`, `"myUdt"`), which is used directly; the raw RID stays on `Member.rid`.
- **Sections.** FB/OB members are tagged with their section (Input/Output/InOut/Static/Temp/Constant); section-header and list-wrapper `<Member>` elements (no RID) are skipped. DB/UDT members are flat (empty section).
- **Nested UDT fields.** When a member is of a UDT/struct, its definition is embedded under the primary part's `<SubParts>` (selected by the member's `SubPartIndex`); it is expanded recursively into `Member.fields`. Library types (e.g. `IEC_TIMER`) are referenced but not inlined, so they expand to no fields. Embedded type definitions are never surfaced as top-level members.

### DB listing
- `datablocks_from_explore()` — parse the DB list from a `0x8A11FFFF` wildcard EXPLORE. `PlcContentInfo` is a *standard* zlib stream (`78 da`) embedded in a larger BLOB, so it is raw-inflated to tolerate the trailing bytes.

## Honest scope / testing

- **Based on the `feat/zlib-preset-dictionaries` branch**, not `master` — it imports the decompressor, so it should merge *after* (or into) #776.
- **Validated on a live S7-1200 G2 (FW V4.1) only.** I no longer have the G1 FW V4.5 unit from #757, but the G2 declares the same dictionaries.

**I/Q/M symbols** (verified against TIA Portal): `input_1 %I0.0`, `output_1 %Q0.1`, `merker_1 %M0.2`, `mtag_byte %MB100`, `mtag_word %MW102`, `mtag_dword %MD104`, `mtag_bool %M108.0`.

**Optimized DB `Data_block_1`** (11 elementary members: Bool/Int/Real, cross-checked against the physical `<Offsets>`), plus a live FB (`FB_Macchina`) with sections + `Array[0..5] of IEC_TIMER` / `Array[0..2] of Bool`, and a DB member of a user UDT `elettrovalvola` that expands into its 6 inner fields (`fc_apertura`, `stato_ev`, …). Tests use real EXPLORE blob fixtures from the G2; 10/10 pass, ruff clean.

This makes the oracle pipeline from #757 (`_build_fdict` / regex `_extract_tags` / symbolic-READ width workaround) obsolete: `<SimpleType>` carries the real M widths directly, and block interfaces resolve straight to clean member names, types, sections and nested UDT fields.

### DIFF
--- a/s7commplus/__init__.py
+++ b/s7commplus/__init__.py
@@ -18,6 +18,14 @@ from .client import S7CommPlusClient as Client
 from .async_client import S7CommPlusAsyncClient as AsyncClient
 from .server import S7CommPlusServer as Server, DataBlock, CPUState
 from .connection import S7CommPlusConnection
+from .tag_browser import (
+    Tag,
+    Member,
+    DataBlock as ExploreDataBlock,
+    tags_from_explore,
+    block_interface_from_explore,
+    datablocks_from_explore,
+)
 
 __all__ = [
     "Client",
@@ -28,4 +36,10 @@ __all__ = [
     "S7CommPlusConnection",
     "decompress_blob",
     "find_and_decompress",
+    "Tag",
+    "Member",
+    "ExploreDataBlock",
+    "tags_from_explore",
+    "block_interface_from_explore",
+    "datablocks_from_explore",
 ]
--- a/s7commplus/tag_browser.py
+++ b/s7commplus/tag_browser.py
@@ -190,33 +190,44 @@ def _member_type(member: ET.Element) -> str:
     return f"0x{rid:08X}"
 
 
-def _enclosing_section(member: ET.Element, parent: dict) -> str:
-    """Walk up to the enclosing ``<Part Kind="XxxSection">`` and return ``Xxx``."""
+def _enclosing_part(member: ET.Element, parent: dict) -> ET.Element | None:
+    """Return the nearest enclosing ``<Part>`` element, or ``None``."""
     node = parent.get(member)
     while node is not None:
-        kind = node.get("Kind", "") if node.tag == "Part" else ""
-        if kind.endswith("Section"):
-            return kind[: -len("Section")]
+        if node.tag == "Part":
+            return node
         node = parent.get(node)
-    return ""
+    return None
 
 
 def parse_block_interface(xml_text: str) -> list[Member]:
     """Parse a decompressed ``<BlockInterface>`` document into :class:`Member`s.
 
-    Works for DB, FB and OB interfaces. Only ``<Member>`` elements that carry a
-    ``RID`` are real variables; the section-header and list-wrapper ``<Member>``
-    elements (which have no ``RID``) are skipped. Members are returned in
-    declaration order, each tagged with its section (Input/Output/... for FB/OB;
-    empty for a DB).
+    Works for DB, FB, OB and UDT interfaces. A member is a real variable only if
+    it carries a ``RID``; the section-header and list-wrapper ``<Member>``
+    elements (no ``RID``) are skipped. When a block uses a UDT, the interface
+    also embeds that UDT's own definition under a nested ``<Part
+    Kind="DataTypeSource">`` — those members describe the *type*, not the block,
+    so they are skipped: a member is kept only when its nearest enclosing
+    ``<Part>`` is the block's primary part (flat DB/UDT) or a ``*Section`` part
+    (FB/OB). Members are returned in declaration order, each tagged with its
+    section (Input/Output/... for FB/OB; empty for DB/UDT).
     """
     root = ET.fromstring(xml_text)
     parent = {child: p for p in root.iter() for child in p}
+    primary = root.find("Part")  # first top-level Part = the block's own source
     members: list[Member] = []
     for m in root.iter("Member"):
         rid_attr = m.get("RID")
         if rid_attr is None:
             continue  # section header or list wrapper, not a variable
+        part = _enclosing_part(m, parent)
+        if part is primary:
+            section = ""
+        elif part is not None and part.get("Kind", "").endswith("Section"):
+            section = part.get("Kind")[: -len("Section")]
+        else:
+            continue  # member of an embedded type definition (nested *Source part)
         members.append(
             Member(
                 name=m.get("Name", ""),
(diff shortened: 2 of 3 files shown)


gijzelaerr (commented):
Nice work — clean module, real hardware fixtures, well-scoped. Three things I noticed:

**1. Off-by-one in `_find_preset_stream`** (`tag_browser.py:262`)

`end = len(data) - 6` should be `len(data) - 5`. The zlib header is 6 bytes (CMF + FLG + 4-byte DICTID), so valid start positions go up to `len(data) - 6` inclusive, but the `while i < end` loop excludes that last position. Unlikely to bite in practice (a stream starting 6 bytes from the end would have no compressed data), but still a real off-by-one.

**2. No cycle guard in recursive UDT expansion** (`tag_browser.py:231`)

`_build_member` follows `SubPartIndex` → `_parse_source_part` → `_build_member` without tracking visited parts. A malformed `BlockInterface` with circular `SubPartIndex` references would hit `RecursionError`. A `seen: set[int]` threaded through the recursion would prevent it. Not a concern with real PLCs, but cheap to add.

**3. DB listing only matches `78 da`** (`tag_browser.py:281`)

`datablocks_from_explore` hardcodes `b"\x78\xda"` (zlib best-compression), but standard zlib streams can use other FLG bytes (`78 9c`, `78 5e`, `78 01`). If a firmware version emits `PlcContentInfo` with default compression the function silently returns `[]`. Scanning for any `0x78` byte and validating the header checksum (CMF*256 + FLG must be divisible by 31) would be more robust.

None of these block merging — just worth fixing while the code is fresh.

tommasofaedo:
Thanks for the careful review — all three fixed in 48c8659:

1. **Off-by-one in `_find_preset_stream`** — `end = len(data) - 5`, so the last valid 6-byte header start position is now scanned.
2. **Cycle guard in recursive UDT expansion** — a `seen` set of `<Part>` identities is threaded through `_parse_source_part`/`_build_member`; a circular `SubPartIndex` now stops instead of hitting `RecursionError`.
3. **DB listing zlib header** — `datablocks_from_explore` now uses a new `_find_zlib_stream` helper that matches any CMF `0x78` with a valid header checksum (`CMF*256 + FLG` divisible by 31) and no preset-dict flag, rather than hardcoding `78 da`, so `78 01`/`5e`/`9c` from other firmwares are tolerated.

10/10 tests still pass, ruff clean.

gijzelaerr:
@tommasofaedo We reviewed this PR — code is clean, tests are excellent, and the dataclass API fits well with the project.

One question: do you think it would be useful to re-export the public types from `s7commplus/__init__.py` so users can do `from s7commplus import Tag, Member` directly? If so, we could add that to this PR. There is a name collision with `DataBlock` from the server module — one option would be to alias it as `ExploreDataBlock`.

What do you think? Happy to hear your perspective on the API surface.

Also: this PR depends on #776 (blob_decompressor), so we'll merge that first.

tommasofaedo:
Happy with the approach — thanks for the clear direction. I've pushed the re-exports to `s7commplus/__init__.py` (commit tommasofaedo/python-snap7@f669720):

```python
from .tag_browser import (
    Tag,
    Member,
    DataBlock as ExploreDataBlock,
    tags_from_explore,
    block_interface_from_explore,
    datablocks_from_explore,
)
```

and added the same six names to `__all__`, keeping the `DataBlock as ExploreDataBlock` alias to avoid colliding with the server emulator's `DataBlock`. So `from s7commplus import Tag, Member` now works.

Since #776 is merged to `master`, happy to retarget this PR's base from `feat/zlib-preset-dictionaries` to `master` (the diff should then reduce to just the tag_browser additions) — let me know if you'd like that or prefer to handle it on merge.

tommasofaedo:
Quick heads-up on the merge path now that #776 is in `master`:

- This branch is currently based on `feat/zlib-preset-dictionaries` — the branch you just merged — so if #780 is left as-is the diff/CI drags in already-merged content and looks noisier than it is. Retargeting the base to `master` collapses it back to just `s7commplus/tag_browser.py`, the `__init__.py` re-exports, and `tests/test_tag_browser.py`.
- If it's easier for you, I'm happy to rebase `followup-tag-browser` onto `master` and force-push so it lands merge-ready with linear history (same as we did on #753) — or leave it and you retarget/rebase on merge, whichever is least work for you.
- Nothing else pending on my side: tests are 10/10, ruff clean, and the re-exports (f669720) are the last change you asked for.

I'll leave the branch untouched until you decide — just say the word and I'll rebase.

gijzelaerr:
@tommasofaedo Great, thanks for adding the re-exports!

Go ahead and rebase onto master — but please don't force-push, just merge master in as a new commit. That way we keep the existing history intact. Once that's done I'll retarget the PR base to master and merge.

tommasofaedo:
Done — merged `master` into the branch as a new merge commit (`a3aba5d`), no force-push, existing history preserved. The merge was clean (no conflicts) and the `tag_browser` tests pass (10/10).

The branch is now up to date with `master` (including #776). Ready for you to retarget the PR base to `master` and merge. Thanks!

Issue #794, opened 2026-08-03
Alarm handling and subscriptions (S7CommPlus)


gijzelaerr:
## Problem

thomas-v2 has full alarm subscription support: `AlarmSubscriptionCreate/Delete`, AS alarm browsing, alarm text retrieval by language. lircy has an Alarming module too. python-snap7 has no alarm support at all.

## Scope

- Alarm subscription create/delete via S7CommPlus CreateObject
- Alarm state browsing
- Alarm text retrieval (multi-language)
- Both sync and async clients

## References

- thomas-v2: `AlarmSubscription` module
- lircy: `Alarming` module

## Priority

Important — table stakes for SCADA/HMI integrations.

PR #815, opened 2026-08-14
fix(tls): allow TLS 1.3 and restrict offered groups in async client

### LINKED ISSUE #814: Inconsistent TLS used in sync vs async client for the S7CommPlus
I was testing and came across that the the SSL context setup is inconsistent, for different reasons.

For the sync client, it is set with a minimum version 1.2:

https://github.com/gijzelaerr/python-snap7/blob/d5d489cbc1e6738305f2b909725e50e48ec34274/s7commplus/connection.py#L1179-L1202

But for the async client, this is done with a minimum and maximum version:

https://github.com/gijzelaerr/python-snap7/blob/d5d489cbc1e6738305f2b909725e50e48ec34274/s7commplus/async_client.py#L240-L259

For the async client, this is related to the commit https://github.com/gijzelaerr/python-snap7/commit/3bbce48b4926b7788bf0ed5e67e18e9d4956efe4 that mentions:

> Setting maximum_version to TLS 1.2 produces a clean TLS 1.2 ClientHello without these extensions.

For the sync client, you mention in https://github.com/gijzelaerr/python-snap7/commit/a186bf5c0047b9c0bb1fd392aadf7072eb2cde17 that:

> Lower the minimum to TLS 1.2 so Python's ssl module auto-negotiates the highest mutually-supported version.

I have tested with a real PLC and found out that setting the `ctx.maximum_version = ssl.TLSVersion.TLSv1_2` causes the setup to fail when the PLC requests TLS1.3 (tested with a S7 1517TF, FW 3.1.5).

@gijzelaerr I'm willing to contribute the change. My idea was to just remove the `maximum_version`. It is possible that the SSL connection will **have to be done depending on the firmware**, if this is problematic.

bvanelli:
Port the client-side TLS compatibility fix to the async client, which was missed when the change landed for the sync connection and server.

- Remove maximum_version=TLSv1_2 pin, which blocked TLS 1.3 PLCs. minimum_version=TLSv1_2 stays as the floor for older firmware.
- Restrict offered groups via _set_s7_groups. Siemens PLCs RST when they see unsupported groups in the ClientHello, regardless of TLS version.

The async context now matches `_setup_ssl_context` in `connection.py` sharing  `_S7_CIPHERS` and `_set_s7_groups` so the cipher string cannot drift between them.

Tested against a S7 1517TF, FW 3.1.5, no special secure connection required, no password (browsing and random read both worked).

Closes #814

### DIFF
--- a/s7commplus/async_client.py
+++ b/s7commplus/async_client.py
@@ -9,7 +9,7 @@ import ssl
 import struct
 from typing import Any, Optional
 
-from .connection import _S7_CIPHERS
+from .connection import _S7_CIPHERS, _set_s7_groups
 from .protocol import (
     DataType,
     ElementID,
@@ -251,9 +251,9 @@ class S7CommPlusAsyncClient:
 
         ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
         ctx.minimum_version = ssl.TLSVersion.TLSv1_2
-        ctx.maximum_version = ssl.TLSVersion.TLSv1_2
 
         ctx.set_ciphers(_S7_CIPHERS)
+        _set_s7_groups(ctx)
         ctx.options |= ssl.OP_NO_TICKET
         ctx.options |= 0x00080000  # SSL_OP_NO_ENCRYPT_THEN_MAC
         ctx.options |= 0x00000001  # SSL_OP_NO_EXTENDED_MASTER_SECRET (OpenSSL 3.0+)


gijzelaerr (approved):

bvanelli:
@gijzelaerr how are you coordinating development and testing for new functionality for this library?

I for example did a working spike for https://github.com/gijzelaerr/python-snap7/issues/794 (reading alarms on a loop), as well as some other small improvements (correct protection handling, raising the right exception when the PLC is password protection). I would like to also clean them up and provide some smaller PRs, as well some sort of contract testing with PLC dumps.

Do you have a Discord/Slack for development things that do not belong on Github?

gijzelaerr:
drop me a email on `<firstname><lastname>@gmail.com` and i'll add you to the slack.

PR #818, opened 2026-08-18
fix(server): encode V2 response IntegrityIds

### LINKED ISSUE #802: S7CommPlus server: encode IntegrityId in V2 response payloads
The S7CommPlus server emulator tracks IntegrityId counters per-client (read/write) but never embeds the IntegrityId bytes in V2+ response payloads. It works today by accident — the client's VLQ parser consumes trailing zeros as IntegrityId=0, which matches the initial counter state — but this breaks as soon as a test exercises multiple reads/writes where the counter advances past 0.

### What needs to happen

- Append the correct IntegrityId VLQ at the end of each V2+ response payload (read counter for read responses, write counter for write responses)
- Validate incoming IntegrityId from the client against the server's counters (optional but useful for catching desync bugs)
- Add tests that do multiple sequential reads/writes on a V2 server and verify the session stays intact

### Context

Discovered during a protocol version audit of the server emulator. V1 is fully supported, V2 is partially supported (TLS works, IntegrityId is fragile), V3 is not supported (#803).

gijzelaerr:
## Summary

- append the current per-client read or write IntegrityId to every established-session V2 response
- preserve the legacy V1 response layout
- preserve the current server's optional TCP-reset behavior after symbolic reads
- cover read/write counter selection and sequential V2 TLS operations

The branch now includes current `master` through a merge commit; its previous conflicts are resolved without rewriting history.

## Verification

- `uv run --frozen --extra test --extra s7commplus pytest -q` (1779 passed, 82 skipped)
- focused V2/TLS/server tests (124 passed)
- `uv build --no-sources`

Fixes #802


gijzelaerr:
@bvanelli Could you review the V2 emulator response behavior here, particularly selecting the read/write counter and appending the IntegrityId without changing the V1 payload layout? The conflict repair also preserves the newer symbolic-read TCP-reset emulation.

The refreshed head passes 1,779 tests, the complete pre-commit suite, and the package build.

gijzelaerr:
@ale-rinaldi you have detailed V2/TLS observations from the 1507S and 1211C. Could you review the server response layout here against your real-PLC captures or traces, particularly whether the read/write IntegrityId selection and appended counter match V2 responses without changing the V1 payload shape?\n\nThis is primarily a capture/protocol review; it does not require exposing project data or pointing a real PLC at the emulator.

s7commplus/server.py
```diff
@@ -619,34 +619,47 @@ def _process_request(
...
         elif function_code == FunctionCode.SET_VAR_SUBSTREAMED:
-            return self._handle_set_var_substreamed(seq_num, req_session_id, request_data), False
+            response = self._handle_set_var_substreamed(seq_num, req_session_id, request_data)
+            rst = False
         else:
-            return self._build_error_response(seq_num, req_session_id, function_code), False
+            response = self._build_error_response(seq_num, req_session_id, function_code)
+            rst = False
+
+        if self._protocol_version >= ProtocolVersion.V2 and req_session_id != 0:
```
bvanelli, line 698: Here, it appends the `integrity_id` to every function. In theory, this could also apply to `_build_error_response` and `_handle_init_ssl`. I'm not sure those use need the integrity id.

gijzelaerr: Good question—the generic placement needed an explicit audit. InitSSL cannot receive this suffix because it runs with a zero request session id, so the active-session guard skips it. An unsupported request made inside an established V2 session does use the write IntegrityId, matching the client, which advances its write counter for every non-read request.

While checking this, I found a nearby real bug: GetVarSubStreamed still emitted its legacy zero field before the generic V2 path appended the current counter, producing two IntegrityIds. The updated head 66afe4f fixes that and adds regressions covering exactly one V2 substreamed IntegrityId, no InitSSL IntegrityId, and the write counter on an in-session error response. The merged tree passes all pinned hooks, source/wheel builds, and the full local suite (2,005 passed, 82 skipped). Thanks for prompting the broader check.

PR #775, opened 2026-07-16
SessionKey handshake for V1-initial S7-1200 PLCs

gijzelaerr:
Replaces #761, updated onto the s7commplus/ package split (#774).

## Summary
Adds the full `session_auth` package for V1-initial S7-1200 PLCs that require a SessionKey handshake before data operations:

- `s7commplus/session_auth/`: public-key-based session authentication (Family 0 key derivation, fingerprint matching, SecurityKeyEncryptedKey blob generation, legitimation challenge solver, HMAC key derivation)
- V3 HMAC framing for post-auth data operations
- TIA-style CreateObject attributes required by V1-initial firmware
- PAOM string stripping in ServerSessionVersion echo
- Post-auth legitimation flow (challenge read + solved blob write)
- `Connection.connect(password=)` parameter for auth flow
- Compatibility gating so TLS/V2 sessions do not enter the V1 SessionKey handshake
- Protocol-version-specific ObjectQualifier framing, preserving fixed-width V1 encoding and VLQ-based V2 encoding
- Correct standard GetVarSubStreamed/SetVariable legitimation framing and IntegrityId placement

## Test plan
- [x] 1759 tests pass, 82 skipped
- [x] Full pre-commit suite passes (mypy, Ruff, formatting, validation)
- [x] Source distribution and wheel build successfully
- [x] Session setup completes on S7-1200 CPU 1212C FW V4.2 (xBiggs hardware testing on #724)
- [x] TLS/V2 compatibility validated on three older-firmware PLCs (FW 2.9.8) by bvanelli
- [ ] Current HEAD validated on a V1-initial S7-1200 by xBiggs
- [ ] Password legitimation retested against hardware (follow-up; not a merge blocker)

## Merge gate
Merge after xBiggs validates the current HEAD on V1-initial S7-1200 hardware. Continue iterating on non-blocking protocol diagnostics in focused follow-ups.

## Known limitations
- Only Family 0 (`01:` prefix) public keys are included
- V254 SystemEvent decoding and error reporting are incomplete; a separate follow-up was requested in https://github.com/gijzelaerr/python-snap7/pull/775#issuecomment-5324346850
- Password legitimation against the affected TLS PLC remains separate follow-up work

PR #821, opened 2026-08-18
feat(server): add V3 HMAC framing

### LINKED ISSUE #803: S7CommPlus server: V3 protocol support (HMAC framing)
The S7CommPlus server emulator has zero V3 logic. Instantiating `S7CommPlusServer(protocol_version=ProtocolVersion.V3)` produces V3-tagged frames without the 32-byte HMAC prefix the client expects, causing the client to misparse every response.

### What V3 requires

- **HMAC prefix:** Every V3 response frame must include a 32-byte HMAC between the frame header and the response data. The client strips this in `send_request` and `collect_explore_frames`.
- **Security key derivation:** The HMAC is computed from session keying material exchanged during CreateObject. The server needs to derive the same symmetric key the client uses for verification.
- **Frame encoding:** `encode_header` already handles V3 header format, but the HMAC insertion and computation are missing.

### What needs to happen

- Implement HMAC computation in the server using the session key material
- Prefix every V3 response frame with the computed HMAC
- Add integration tests: V3 client connecting to V3 server, performing reads/writes/explore
- Consider whether V3 TLS is also needed (V3 is the TLS-required generation on real S7-1500 hardware)

### Context

Discovered during a protocol version audit. V1 is fully supported, V2 is partially supported (#802), V3 is absent.

gijzelaerr:
## Summary

- verify HMAC-SHA256 on incoming V3 server requests and sign every V3 response
- validate V3 response HMACs in the sync client, including reassembled Explore frames
- carry V3 response IntegrityIds without exposing them to response parsers
- add end-to-end SessionKey tests covering connect, read, write, and Explore

The emulator accepts an explicit 24-byte-or-longer `session_key`. It cannot derive the client random key from material encrypted for a real Siemens PLC without the corresponding private key, so this gives tests and emulator users matching negotiated key material.

## Dependency

#775 has merged. This branch now includes current `master` via a merge commit and targets `master` directly.

## Validation

- `uv run --frozen pytest -q` (1824 passed, 78 skipped)
- focused V3/server/TLS tests (119 passed)
- `uv build --no-sources`

Fixes #803


gijzelaerr:
@bvanelli Could you review the V3 framing boundary here, particularly the HMAC input, response IntegrityId placement, and the emulator's explicit session-key setup? This is emulator-focused, so I am looking for a protocol review rather than another PLC capture.

The refreshed head is based on current `master` and passes 1,824 tests plus the full pre-commit and build checks.

tests/test_s7_server.py
```diff
@@ -419,9 +423,28 @@ def test_session_key_server_construction(self) -> None:
...
         assert srv._session_challenge == TEST_CHALLENGE
+        assert srv._session_key == TEST_SESSION_KEY
+
+    def test_short_session_key_rejected(self) -> None:
+        with pytest.raises(ValueError, match="at least 24 bytes"):
+            S7CommPlusServer(session_key=b"too short")
+
+    def test_v3_hmac_verification(self) -> None:
+        application_data = b"S7CommPlus request"
+        digest = hmac.new(TEST_SESSION_KEY, application_data, hashlib.sha256).digest()
```
bvanelli, line 438: I know it's not relevant, and it is personal preference, but this test reinforces what `_verify_v3_hmac` already does. For a real immutable test I would take a [Wikipedia example](https://en.wikipedia.org/wiki/HMAC), because it is verifiable.

gijzelaerr: Good point about making fixtures independently inspectable. This test does use Python stdlib hmac.new to produce the expected digest, so it is not only comparing our two verifier implementations to each other. I will keep the existing fixture in this focused PR; a published fixed vector would still be a useful follow-up if we expand the low-level HMAC tests.

gijzelaerr:
@bvanelli Following your earlier feedback and recent framing/authentication contributions, could you take another look at the current V3 HMAC server implementation?

bvanelli (approved):
LGTM. I can hopefully test the framing with hardware in the future.

PR #823, opened 2026-08-18
feat(client): add request rate limiting

### LINKED ISSUE #628: Request rate limiting to protect PLC resources
## Summary

Add configurable request rate limiting to prevent overwhelming PLCs with too many requests, especially older/smaller models with limited processing capacity.

## Background

Siemens S7-200 and S7-300 PLCs have limited communication resources. Sending requests too fast can cause the PLC to drop connections, queue requests, or even affect scan cycle times. A built-in rate limiter would protect against this, especially in polling loops.

## What needs to be done

- [ ] Implement a configurable rate limiter (max requests per second)
- [ ] Support both fixed-rate and token-bucket algorithms
- [ ] Apply rate limiting transparently to all client operations
- [ ] Add per-connection and optionally per-PLC-model rate limits
- [ ] Add configurable behavior when limit is hit: block (wait), raise exception, or drop
- [ ] Make it opt-in (disabled by default)
- [ ] Add tests
- [ ] Add documentation

## API Design Ideas

```python
client = snap7.Client(
    max_requests_per_second=10,  # 0 for unlimited (default)
    rate_limit_behavior="block",  # "block", "raise", or "drop"
)

client.connect("192.168.1.10", 0, 1)

# Operations are automatically throttled
for i in range(1000):
    data = client.db_read(1, 0, 4)  # Will not exceed 10 req/s
```

## Notes

- Different PLC models have different limits:
  - S7-200: ~10 requests/second
  - S7-300: ~20-50 requests/second
  - S7-400: ~100+ requests/second
  - S7-1200/1500: ~100+ requests/second
- Could provide sensible defaults per PLC model (auto-detected from CPU info)
- Rate limiting should account for multi-variable operations (one multi-read = one request, not N)
- Should work with both sync and future async clients

gijzelaerr:
## Summary

- add opt-in per-client request limiting at the outbound S7 PDU boundary
- support evenly spaced fixed-rate and burst-capable token-bucket algorithms
- support blocking and immediate rejection when the limit is reached
- cover both synchronous and asynchronous clients
- document configuration and add deterministic limiter tests

A multi-variable PDU counts as one request, while chunked and multi-packet operations count each actual PDU. Model-specific defaults are intentionally not inferred: the limiter is disabled by default and users choose a rate appropriate for their PLC and workload.

Silent drops are unsafe at the PDU boundary because the surrounding client operation would wait for a response to a request that was never sent. The non-blocking behavior therefore raises `S7RateLimitError` immediately.

The branch includes current `master` through a merge commit.

## Validation

- `uv run --frozen --extra test --extra s7commplus pytest -q` (1,902 passed, 78 skipped)
- focused limiter/client tests (163 passed, 1 skipped)
- `uv build --no-sources`

Fixes #628


gijzelaerr:
@NiklasReisser Given your recent work around server behavior and application-facing semantics, could you review the request-rate-limiting API and concurrency behavior here?

snap7/rate_limiter.py
```diff
@@ -0,0 +1,110 @@
...
+            delay = queued_delay + ((1.0 - self._tokens) / self.rate)
+            if self.behavior != "block":
+                self._reject()
+            self._tokens = 0.0
+            self._last_refill = now + delay
+            return delay
+
+    def _reject(self) -> None:
+        dropped = self.behavior == "drop"
+        action = "dropped" if dropped else "rejected"
```
NiklasReisser: I don't understand the difference between drop and reject. Both raise, just the flag on the exception is different.

snap7/rate_limiter.py
```diff
@@ -0,0 +1,110 @@
...
+            if now > self._last_refill:
+                elapsed = now - self._last_refill
+                self._tokens = min(float(self.burst_capacity), self._tokens + elapsed * self.rate)
+                self._last_refill = now
+
+            if self._tokens >= 1.0:
+                self._tokens -= 1.0
+                return 0.0
+
+            queued_delay = max(0.0, self._last_refill - now)
```
NiklasReisser: Earlier you set `_last_refill` to `now`, if `_last_refill` is in the past. Thus, is this `max()` required? `last_refill` will be at least equal to `now`.

snap7/rate_limiter.py
```diff
@@ -0,0 +1,110 @@
...
+            if self.algorithm == "fixed":
+                delay = max(0.0, self._next_request - now)
+                if delay > 0 and self.behavior != "block":
+                    self._reject()
+                slot = now + delay
+                self._next_request = slot + (1.0 / self.rate)
+                return delay
+
+            # Refill only through the current time. A future _last_refill
+            # represents tokens already reserved by blocking callers.
```
NiklasReisser: Might be a stupid question. But to quote the docs: `The Client class is not thread-safe. Concurrent calls from multiple threads on the same Client instance will corrupt the TCP connection state and cause unpredictable errors.` 
So, how can there be blocking callers?

README.rst
```diff
@@ -47,6 +47,26 @@ Connect to any S7 PLC::
...
+       max_requests_per_second=10,
+       rate_limit_algorithm="fixed",       # or "token_bucket"
+       rate_limit_behavior="block",         # or "raise" / "drop"
+   )
+
+``fixed`` spaces requests evenly. ``token_bucket`` permits a burst (one second
+of requests by default, configurable with ``rate_limit_burst``) and then
+refills at the configured rate. The default rate is ``0``, which disables the
+limiter. ``raise`` and ``drop`` both raise ``S7RateLimitError`` immediately;
+for ``drop``, its ``dropped`` attribute is true. This avoids waiting for a PLC
```
NiklasReisser: But where is the actual difference in behavior? Both raise immediately.

snap7/rate_limiter.py
```diff
@@ -0,0 +1,110 @@
...
+    A rate of zero disables limiting.
+    """
+
+    def __init__(
+        self,
+        max_requests_per_second: float = 0,
+        *,
+        algorithm: RateLimitAlgorithm = "fixed",
+        behavior: RateLimitBehavior = "block",
+        burst_capacity: int | None = None,
```
NiklasReisser, line 31: To test my own understanding: Is there any point in a `burst_capacity` smaller then `max_requests_per_second`?

snap7/rate_limiter.py
gijzelaerr: Agreed. There was no useful behavioral distinction, and silently dropping at this PDU boundary would make the surrounding operation wait for a response forever. I removed drop; the supported behaviors are now block and raise.

snap7/rate_limiter.py
gijzelaerr: Correct—the max was redundant because the refill branch establishes the invariant that last_refill is at least now. Removed in cd631ed.

snap7/rate_limiter.py
gijzelaerr: The lock protects only limiter accounting; it does not make Client thread-safe. A future last_refill can represent capacity reserved by another direct acquire call before that caller sleeps. I clarified this scope and added a test for distinct queued reservations.

README.rst
gijzelaerr: Agreed; there was none beyond exception metadata. I removed drop from the API and documentation and kept raise as the sole non-blocking behavior.

snap7/rate_limiter.py
gijzelaerr: Yes. The rate controls token refill while burst_capacity independently caps how many requests may leave at once. For example, rate 10 with capacity 1 permits 10 requests per second but no initial burst. This is now documented and tested.

gijzelaerr:
@amorelettronico you previously tested the pure-Python client on CPU 414-5H and CPU416-2 hardware. Could you run a small real-PLC smoke test of the opt-in rate limiter on an S7-400 when convenient?\n\nThe useful check would be a low fixed rate (for example 5 requests/second), confirming reads still succeed and are spaced as configured, followed by disabling the limiter and confirming normal behavior returns. No production load test or sensitive PLC data is needed.

amorelettronico:
Hi @gijzelaerr 

### Hardware smoke test — S7-416-2 DP

Tested PR #823 against a Siemens S7-416-2 DP using DB15 as a dedicated test DB.

**Configuration**

* Test DB: `DB15`
* Rate limit: `5 requests/sec`
* Expected interval: `200 ms`
* Algorithm: `fixed`
* Behavior: `block`

 **DB read**

* Successfully read DB15.
* Returned values matched the expected test data.

 **Rate limiter — 5 requests/sec**

* 15 consecutive DB reads completed successfully.
* Measured intervals:

```text
0.200, 0.201, 0.199, 0.199, 0.202,
0.198, 0.199, 0.202, 0.201, 0.200,
0.198, 0.200, 0.201, 0.200 sec
```

* Average interval: **0.200 sec**


**Limiter disabled**

* With `max_requests_per_second=0`, reads completed successfully without the intentional 200 ms delay.
* Measured intervals were approximately **6–15 ms**, confirming that the rate limiter was no longer imposing the 5 req/sec limit.

### Conclusion

The rate limiter behaves as expected on the S7-416-2 DP with a real PLC connection.

**PR #823 smoke test: PASS**.

gijzelaerr:
Thanks @amorelettronico — this covers the requested hardware smoke test: successful reads, approximately 200 ms spacing at 5 requests/sec, and normal timing with the limiter disabled.

@NiklasReisser the review findings were addressed in cd631ed and replied to inline. All current checks pass, and the S7-416-2 DP hardware test now passes too. Could you take another look and submit a review? The existing review request is still pending.

NiklasReisser (approved):

Issue #765, opened 2026-07-10
Regression: legacy client connect ignores set_connection_type in TSAP composition (sync + async)


yangyachao:
## Summary

In python-snap7 3.0.0, the legacy pure-Python client path appears to ignore set_connection_type during connect TSAP composition.

As a result, PG/OP separation behavior is lost in real PLC scenarios where multi-client coexistence depends on connection type.

This affects both sync and async legacy clients.

## Environment

- OS: Ubuntu 24.04
- Python: 3.14.4
- PLC: Siemens S7 (TCP 102), observed on S7-200 SMART behavior tests
- Affected: python-snap7 3.0.0
- Baseline (expected field behavior): python-snap7 2.1.0

## Minimal Reproduction

1. Holder client:
   - call set_connection_type(2)
   - connect and keep connection alive

2. Probe client:
   - test ct=1 (PG): connect + DB read
   - test ct=2 (OP): connect + DB read

3. Repeat with holder ct=1 and holder ct=2

## Actual

With holder on 3.0.0 legacy path:

- PG probe fails (timeout/reset)
- OP probe succeeds

Observed behavior is nearly identical regardless of holder ct=1 or ct=2, suggesting connection type is not reflected in called TSAP during connect.

## Expected

set_connection_type(2) should produce OP behavior and preserve PG/OP coexistence semantics, consistent with observed 2.1.0 field behavior.

## Code Evidence

Current code suggests connect overwrites TSAP without using connection_type:

- snap7/client_base.py:141: set_connection_type only stores self.connection_type
- snap7/client.py:585: remote_tsap is computed as 0x0100 | (rack << 5) | slot
- snap7/async_client.py:397: same fixed composition in async connect
- snap7/connection.py:320: remote_tsap is directly encoded into COTP called TSAP

## Local Validation Patch

Using connection_type in TSAP composition during connect restores expected field behavior in our test setup:

self.remote_tsap = (self.connection_type << 8) | (rack << 5) | slot

After this local change, a 3.0.0 holder with ct=2 allows both PG and OP probes to connect/read in our field test.

## Request

Could maintainers confirm whether this was an intentional behavior change or a regression?

If this is a regression, I can provide:

- a compact runnable reproducer script
- packet/log captures from 2.1.0 vs 3.0.0 for comparison

gijzelaerr:
[...]

PR #824, opened 2026-08-18
feat(ppi): add S7-200 serial transport

### LINKED ISSUE #614: PPI (Point-to-Point Interface) protocol support
## Summary

Implement PPI protocol support for communicating with Siemens S7-200 series PLCs.

## Background

PPI (Point-to-Point Interface) is the native communication protocol for Siemens S7-200 series PLCs. While S7-200 is an older product line, it remains widely deployed in legacy industrial installations worldwide. PPI uses serial communication (RS-485) and has a different frame format than the TCP/IP-based S7 protocol.

## What needs to be done

- [ ] Implement PPI frame encoding/decoding (SD1, SD2, SD3, SC, ED frame types)
- [ ] Implement PPI address handling (station addresses 0-126)
- [ ] Implement serial port communication layer (via pyserial)
- [ ] Implement PPI read/write operations for S7-200 memory areas (V, I, Q, M, S, T, C, SM, AI, AQ)
- [ ] Implement token passing mechanism (PPI is a token-ring protocol)
- [ ] Consider PPI over TCP support (for serial-to-Ethernet converters)
- [ ] Add optional dependency on `pyserial`
- [ ] Add tests (mock serial port or PPI-over-TCP)
- [ ] Add documentation and examples

## Notes

- PPI is half-duplex on RS-485
- Default baud rate: 9600 (configurable up to 187.5k)
- S7-200 supports up to 4 master connections
- Many S7-200 installations use serial-to-Ethernet converters, so PPI-over-TCP would be valuable

gijzelaerr:
## Summary

- add SD1, SD2, SD3, and short-confirmation PPI frame codecs
- add an optional pyserial transport implementing the S7-200 request, acknowledgement, poll, and response exchange
- add a PPI client with S7 PDU negotiation and read/write support for S, SM, AI, AQ, I, Q, M, V, counter, and timer areas
- document installation and usage, and cover framing, retries, area mappings, and client behavior with tests

## Initial scope

This is the first point-to-point PC-master/single-slave milestone for #614. Multi-master token passing and PPI-over-TCP are intentionally deferred until protocol traces or suitable hardware are available. The implementation is marked experimental and still needs validation against a real S7-200 PLC.

The wire exchange and area mappings follow the mature libnodave implementation:
https://github.com/netdata-be/libnodave/blob/master/nodave.c#L4077-L4220

The branch now includes current `master` through a merge commit.

## Validation

- `uv run --frozen --extra test --extra s7commplus --extra ppi pytest -q` (1799 passed, 82 skipped)
- focused PPI/protocol/datatype tests (129 passed)
- `uv build --no-sources`

Fixes #614


gijzelaerr:
@Poseidonas You recently contributed protocol/server fixes and careful regression coverage. Could you give this PPI transport implementation a code and test review?

Poseidonas:
Had a look. The framing and the client layer hold up well under abuse: I fuzzed `decode_frame` for 900k mutated frames across the four delimiters and got no exception outside `S7ProtocolError`, so the length, checksum and delimiter checks are doing their job. Round trips, the address bounds on `read_area`/`write_area`, the 246-byte PDU cap and the E5 polling flow all behave as documented when I drive them through a scripted serial port.

One thing I would raise before this lands.

**Line corruption is not retried, only timeouts are.** `exchange` wraps the acknowledgement read in `except S7ConnectionError`, so a silent slave is retried. Everything `decode_frame` rejects raises `S7ProtocolError`, which propagates straight out of the retry loop, and the poll loop has no `except` at all. Driving the transport with `retries=3`:

| line condition | request attempts |
| --- | --- |
| slave never answers | 3 |
| stray byte before the E5 | 1 |
| corrupted checksum in the response | 1 |
| corrupted byte in the payload | 1 |
| broken end delimiter | 1 |

The last row is the one that bothers me: a single flipped bit on a 9600 8E1 line ends the exchange, and in the case where noise is followed by a perfectly good frame the good frame is never read, because `_read_frame` has already raised on the first byte. On RS-485 that is the failure `retries` exists to absorb. Treating a decode failure as retryable, or hunting for the next start delimiter before giving up, would make the retry count mean what it says.

Two smaller notes, neither introduced here.

`build_item_spec` has a lower bound on `start` but no upper one, so an address at or above 2097152 wraps silently: `start=2097152` and `start=0` produce the same `000000` on the wire, and `start=2**31` surfaces a raw `struct.error` rather than an S7 error. That is shared code and predates this PR, but `write_area` is a new public entry point to it and already validates `count` and the data length, so a bound on `start` would fit naturally alongside them. A silently misdirected write is an unpleasant failure mode on a PLC.

`PPITransport.exchange` takes a lock, which reads as a promise of thread safety, but the sequence counter is advanced in `build_*_request` before the lock and validated in `PPIClient._exchange` after it. Six concurrent `read_area` calls on one client gave me five spurious `S7StalePacketError`. I checked and `Client._send_receive` behaves the same way, so this is not something this PR introduces, and it may well be out of scope, but the lock does invite the assumption.

What I could not check is the wire dialogue itself. I have no S7-200, so the SD2/E5/SD1/SD2 sequence, the `0x6C` request control and the `0x5C`/`0x7C` poll toggle are taken on trust from libnodave; everything above is from running the code, not from hardware.

gijzelaerr:
Addressed the serial robustness findings in 992cbba.

- acknowledgement decode failures now participate in the configured retry loop;
- corrupt frames encountered while polling are skipped within the bounded poll window, allowing a following valid frame to be consumed;
- byte-addressed areas reject offsets above the 21-bit byte range before they can wrap into the 24-bit wire address (timers/counters retain their full 24-bit item-index range);
- request construction, transport exchange, parsing, and sequence validation now share one client lock, so concurrent calls cannot advance the sequence outside the serialized exchange.

Added regression coverage for corrupt acknowledgements, noise before a valid response, and unencodable write/read addresses. Validation: 1,804 passed, 82 skipped; full pre-commit suite and source/wheel build pass.

gijzelaerr:
@yangyachao you recently tested python-snap7 against a real S7-200 SMART setup in #765. Before asking you to test this PR: does that PLC expose a physical serial/PPI interface, and do you have an RS-485 adapter available?\n\nThis PR implements serial PPI, so Ethernet/port-102 testing would not validate the missing wire path. A quick yes/no on hardware availability is all we need for now.

gijzelaerr:
Merged final current `master` into this branch without rebasing. The final PR diff remains limited to the PPI implementation/example/tests/docs and the optional `pyserial` dependency/lock entry; there are no workflow or executable-install changes. Local validation on head `79a15b3`: all pre-commit hooks passed, 2106 tests passed (82 skipped), and source/wheel builds passed. Keeping this open: the remaining acceptance blocker is a real S7-200 serial/PPI wire test with an RS-485 adapter; Ethernet testing is not sufficient.

gijzelaerr:
Merged current master into the branch without rebasing and resolved the README conflict by retaining the current 3.2/classic-S7 text plus the experimental PPI section. The final PR delta remains the PPI implementation/docs/tests/optional pyserial dependency, with three one-character RST underline fixes needed because the inherited v6 merge-conflict hook treats an exact seven-equals heading as a conflict marker.

Validation on `8abaacd`:

- 1,412 passed, 49 skipped
- all pre-commit hooks passed, including mypy and Ruff
- Sphinx passed with warnings as errors
- source and wheel builds passed offline
- Twine metadata checks passed
- the complete hosted OS/Python, build, CodeQL, pre-commit, and Read the Docs matrix passed

This is conflict-free and locally healthy again, but I am not merging it yet: the PR still lacks the explicitly required real S7-200 serial/PPI test over an RS-485 adapter. Ethernet testing does not validate this transport.

Issue #837, opened 2026-08-19
Add a typed symbolic tag catalog and read_tag/write_tag API


gijzelaerr:
## Problem

The current symbolic API requires callers to supply raw access-area IDs, LID paths, SymbolCRC values and byte payloads. Browsing can discover symbols, but there is no stable public path from a symbol name to a typed read or write.

Other active implementations provide cached symbol resolution, typed values, batch access and automatic SymbolCRC recovery.

## Scope

- Introduce a public symbolic tag descriptor preserving name, access path, datatype, array metadata and SymbolCRC.
- Build a catalog from the existing browse/type-info machinery.
- Add `resolve_tag`, `read_tag`, `write_tag`, `read_tags` and `write_tags`.
- Encode writes according to the resolved PValue datatype instead of always using raw BLOB semantics.
- Detect SymbolCRC mismatch, invalidate the affected catalog and re-resolve once for safe read operations.
- Preserve the existing low-level symbolic methods.

## Acceptance criteria

- A caller can read and write an optimized DB tag by name without manually handling LIDs.
- Scalar and array metadata survive browsing and resolution.
- Batched results retain per-item success/error information.
- Writes are never automatically retried after an ambiguous transport failure.
- Program-layout change tests demonstrate deterministic CRC invalidation and re-resolution.

## Out of scope

- Subscription lifecycle
- Alarm text/catalog support (#794)
- Persistent on-disk catalogs in the first PR

Issue #836, opened 2026-08-19
Define and implement the supported async S7CommPlus authentication boundary


gijzelaerr:
## Problem

`AsyncClient` currently resembles the synchronous TLS path but lacks V1 SessionKey authentication, legacy HMAC framing, session activation, post-auth legitimation, key renewal and a password argument on `connect()`.

Package documentation broadly claims V1/V2/V3 support without making this async limitation clear. Maintaining two copied protocol state machines is also causing correctness drift.

## Scope

First decide and document one explicit support boundary:

1. Bring async legacy authentication to parity, preferably by sharing protocol codecs/state transitions with the sync connection; or
2. Deliberately support TLS-only async operation and fail early with a clear error on legacy SessionKey PLCs.

In either case:

- Share authentication/frame parsing logic where practical.
- Align connection/session-ready semantics with the sync client.
- Add an explicit compatibility matrix to API documentation.
- Prevent unsupported firmware from yielding a misleading connected client.

## Acceptance criteria

- Async behavior for V1, V2 and V3 is explicitly documented and enforced.
- Supported paths have end-to-end setup/authentication tests.
- Unsupported paths fail during connect with an actionable message.
- Common framing and authentication rules are not independently reimplemented without parity tests.
- Future sync authentication fixes have an obvious shared or mirrored async test location.

PR #839, opened 2026-08-19
docs: correct S7CommPlus support guidance

gijzelaerr:
## Summary

- document the legacy V1 SessionKey path instead of claiming a PUT/GET fallback
- clarify the synchronous and asynchronous authentication boundaries
- replace the nonexistent S7CommPlus `read_tag` example with the current `read_symbolic` API
- distinguish classic S7 symbol/layout limitations from S7CommPlus browsing capabilities
- correct the remaining `browse()` docstrings and runtime warning so they no longer advertise nonexistent APIs or fallback behavior

Related to #710, #836, and #837.

## Validation

- `uv run --frozen sphinx-build -E -N -b html doc/ doc/_build -W`
- `uv run --frozen pytest` (1821 passed, 78 skipped)
- `uv build --no-sources`


gijzelaerr:
@tommasofaedo @bvanelli You have both recently validated and contributed to the S7CommPlus implementation. Could one of you review this support-guidance wording against the functionality available today?

tommasofaedo:
Reviewed the wording against the current code at `631f6ac` — it's accurate. Confirmations:

- The old `client.read_tag(tag)` / `write_tag` example did call nonexistent methods on the S7CommPlus `Client`; `read_tag`/`write_tag`/`Tag.from_access_string` live on the classic `s7.Client`. The replacement with `read_symbolic`/`write_symbolic` matches the real signatures (`read_symbolic(access_area, lids, symbol_crc=0)`), and `access_sequence[0]` = access area / `[1:]` = LIDs is correct.
- Sync `connect(password=...)` applying to either the V1 SessionKey `_post_auth_legitimation` or the TLS legitimation (`authenticate()` when `tls_active and not requires_substreamed`) is accurate; `authenticate` is TLS-only on both clients; the async client has no `password` on `connect` and only supports the TLS legitimation path — all correct.
- The V1/SessionKey note (no PUT/GET fallback) and the rewritten `connection.py` warning match the code.
- The `limitations.rst` reframing (classic S7 vs S7CommPlus `browse()`) matches what I validated live on an S7-1500 (G2): browse reads the symbol tree and reconstructs type info + optimized symbolic access paths.

One runtime-dependent line worth a sanity check on hardware: the `struct.unpack(">f", raw)` example assumes `read_symbolic` returns the bare big-endian value bytes — consistent with my experience, just flagging it as the one behavioral (vs signature-level) assumption. LGTM.

tommasofaedo:
One more wording point I can confirm: the "auto-detects whether the TLS PLC uses the legacy SHA-1 challenge or the newer AES-256-CBC challenge" description matches the code — `authenticate()` builds the response via `build_new_response` (AES-256-CBC) or `build_legacy_response` (SHA-1 XOR) from `legitimation.py`, so the two challenge types named in the doc are the real ones.

PR #847, opened 2026-08-25
fix(server): treat a client disconnect as a normal end, not an error

### LINKED ISSUE #846: Pure-Python server logs a normal client disconnect as an error
### Describe the bug

The pure-Python server logs an error when a client disconnects normally. Using the library's own client:

```python
client = snap7.client.Client()
client.connect("127.0.0.1", 0, 0, 11102)
client.db_read(1, 0, 4)
client.disconnect()
```

```
ERROR  Error handling client ('127.0.0.1', 58143): Expected COTP DT, got 0x80
```

`0x80` is `COTP_DR`, defined in `snap7/connection.py` — `Client.disconnect()` calls `_send_cotp_disconnect()`, which sends it. `ServerISOConnection.receive_data()` accepts only `COTP_DT` and raises `S7ConnectionError` for anything else, so the disconnect reaches the general handler in `_handle_client()` and is logged at error level. `COTP_DR` and `COTP_DC` are defined in the server too, but nothing on the receiving side uses them.

A second case: a peer that goes away before the ISO handshake finishes — a port scan, a health check, a cancelled connect — is logged by `accept_connection()` as `Error accepting ISO connection: Connection closed by peer`.

### To Reproduce

Start a server, then leave in each of these ways, watching the log:

| how the client leaves | logged as |
| --- | --- |
| `Client.disconnect()` after a request | `ERROR Expected COTP DT, got 0x80` |
| `Client.disconnect()` without a request | `ERROR Expected COTP DT, got 0x80` |
| plain TCP close, no ISO handshake | `ERROR Error accepting ISO connection` |
| abrupt reset | `ERROR Error accepting ISO connection` |

### Expected behavior

A disconnect request is how a client is supposed to say goodbye, so it should end the connection normally: confirm it with a `COTP_DC` and take the path `_handle_client()` already has for `ConnectionResetError`, which logs `Client ... disconnected` at info. A peer leaving before the handshake completes is routine and does not need error level either.

Anyone monitoring these servers currently sees an error for every clean disconnect, which buries the errors that matter.

### Environment

python-snap7 master (8838aa9), Python 3.14, macOS. Reproduced against the pure-Python server with the library's own client and with a raw socket.

I have a fix with tests if it is wanted.

Poseidonas:
Fixes #846.

`Client.disconnect()` sends a COTP Disconnect Request, which `receive_data()` rejected as an unexpected PDU, so an ordinary goodbye from the library's own client was logged as an error:

```
ERROR  Error handling client ('127.0.0.1', 58143): Expected COTP DT, got 0x80
```

`COTP_DR` and `COTP_DC` were already defined in `ServerISOConnection` but never used on the receiving side. A DR is now confirmed with a DC and ends the connection through the path `_handle_client()` already treats as a normal disconnect. Sending the confirmation is best effort — a client that closes right after the request may already be gone, which is what the library's own client does.

`accept_connection()` logged a peer that leaves before the handshake completes the same way; that is routine, so it is reported at info level now.

**Measured**, before and after, on the four ways a client can leave — a clean disconnect after a request, a clean disconnect without one, a plain TCP close and an abrupt reset. All four logged an error before; none do now, and the normal disconnect reads:

```
DEBUG  Received COTP DR from client
INFO   Client ('127.0.0.1', 58156) disconnected
```

Three tests cover it: the DC framing, that a DR ends the connection and is confirmed, and that it still ends cleanly when the confirmation cannot be sent. Each fails without the change. `pytest` is 1774 passed, and `mypy` and `ruff` report the same counts as master.

One thing I did not touch: `snap7/connection.py` and `snap7/async_client.py` have the same `Expected COTP DT` check on the client side, so a device that sends a DR would be reported as an error there too. I have no hardware to see whether a real CPU does that, so I left it alone rather than guess.

### DIFF
--- a/snap7/server/__init__.py
+++ b/snap7/server/__init__.py
@@ -625,7 +625,6 @@ class Server:
 
             # Handle ISO connection setup
             if not connection.accept_connection():
-                logger.warning(f"Failed to establish ISO connection with {address}")
                 return
 
             logger.info(f"ISO connection established with {address}")
@@ -2607,9 +2606,6 @@ class ServerISOConnection:
             return True
 
         except (ConnectionResetError, ConnectionAbortedError, TimeoutError) as e:
-            # A peer that goes away before the ISO handshake completes is
-            # routine - port scans, health checks, a cancelled connect - and
-            # says nothing about this server.
             logger.info(f"Peer left before the ISO connection was established: {e}")
             return False
         except Exception as e:
@@ -2645,8 +2641,6 @@ class ServerISOConnection:
             pdu_len, pdu_type, eot_num = struct.unpack(">BBB", payload[:3])
 
             if pdu_type == self.COTP_DR:
-                # The peer is closing the connection the way ISO 8073 says to;
-                # confirm it and let the caller treat this as a normal end.
                 logger.debug("Received COTP DR from client")
                 try:
                     self.socket.sendall(self._build_tpkt(self._build_cotp_dc()))
--- a/snap7/server/__init__.py
+++ b/snap7/server/__init__.py
@@ -2606,6 +2606,12 @@ class ServerISOConnection:
             logger.debug("ISO connection established")
             return True
 
+        except (ConnectionResetError, ConnectionAbortedError, TimeoutError) as e:
+            # A peer that goes away before the ISO handshake completes is
+            # routine - port scans, health checks, a cancelled connect - and
+            # says nothing about this server.
+            logger.info(f"Peer left before the ISO connection was established: {e}")
+            return False
         except Exception as e:
             logger.error(f"Error accepting ISO connection: {e}")
             return False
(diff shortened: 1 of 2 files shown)


gijzelaerr (changes requested):
There is one remaining noisy path in the same scenario. `accept_connection()` now logs an early peer departure at info and returns `False`, but `_handle_client()` immediately turns that result into `logger.warning("Failed to establish ISO connection ...")`. Port scans, health checks, and cancelled handshakes therefore still produce a warning even though the new comment says they are routine.

Please change that outer log to the intended normal severity (or return enough information to distinguish routine departure from a malformed handshake), and add a log-level regression test for the complete `_handle_client()` path. The existing tests call `receive_data()` directly and do not cover the remaining warning.

Poseidonas:
Done — the outer `logger.warning` is removed rather than lowered.

Every path that makes `accept_connection()` return `False` already logs on its own, at the severity it deserves: a routine departure at info, an invalid TPKT version or length at error, a short or wrong-typed COTP CR at error in `_parse_cotp_cr()`, and a partial frame past the deadline through the generic handler. The outer log was therefore a duplicate in every case, so no extra return value is needed to tell the two apart — the distinction is already in the log.

Added `TestHandshakeLogging` with four tests that go through the complete `_handle_client()` path over real sockets, not `receive_data()` directly: a connect-and-close (port scan), a partial TPKT then close, a malformed TPKT, and a full connect/disconnect with the library's own `Client`. Each asserts no record at WARNING or above; the malformed one also asserts the ERROR is still there, so the change does not silence a real problem. Three of the four fail against the previous code. The class binds to port 0 and reads the assigned port back, following `test_max_clients_is_enforced`, so it does not collide when tests run in parallel.

Unrelated, noticed while running the checks: `mypy` reports `snap7/client.py:1404: Redundant cast to "list[dict[str, Any]]"` on master as well. The `cast` after the `hasattr(items[0], "Area")` branch is no longer needed since the early return already narrows the type; dropping it leaves mypy clean with no other change. Happy to open a separate one-line PR for it if that is useful — it seemed out of place in this one.

gijzelaerr (approved):
The duplicate outer warning is gone, malformed handshakes still retain their error log, and the new end-to-end socket tests cover all four requested paths. I also ran the complete server test module locally: 72 passed.

PR #820, opened 2026-08-18
Add S7CommPlus alarm handling and subscriptions

gijzelaerr:
## Summary

- add sync and async alarm subscription create/delete APIs
- add active-alarm browsing with multilingual text filtering
- decode unsolicited alarm notifications into public `Alarm`, `AlarmText`, and `AlarmNotification` models
- default subscriptions to a notification credit limit of 10, matching the working S7-1500 reference trace

## Testing

- `uv run --frozen pytest` (1650 passed, 82 skipped)
- `uv run --frozen pre-commit run --all-files`
- `uv build --no-sources`

The wire-level tests cover subscription filters, active alarm state, coming timestamps, multilingual text payloads, sync/async client APIs, and notification framing. Real PLC notification delivery still needs validation because alarm object contents and delivery behavior vary by firmware.

Fixes #794

PR #848, opened 2026-08-26
fix: Align all 0x34 and 0x36 framing bytes

bvanelli:
## Changes

Aligns the transmission with the received bytes when it comes to the transport flags. The C# library has the following definitions:

| reference request class | TransportFlags |
| --- | --- |
| `InitSslRequest.cs:23` | `0x30` |
| `CreateObjectRequest.cs:23,57` | `0x36` |
| `DeleteObjectRequest.cs:23` | `0x34` |
| `ExploreRequest.cs:24` | `0x34` (with a `// or 0x36???` comment) |
| `GetMultiVariablesRequest.cs:24` | `0x34` |
| `GetVarSubstreamedRequest.cs:23` | `0x34` |
| `SetMultiVariablesRequest.cs:24` | `0x34` |
| `SetVariableRequest.cs:23` | `0x34` |

I aligned based on the testing with a S71500 PLC, and it seems to accept the framing 0x34 or 0x36 in _most_ cases, but sometimes the wrong frame leads to a connection reset. Here are the sent and received frames:

| request | tx before | tx now | rx |
| --- | --- | --- | --- |
| InitSsl | `0x30` | `0x30` | `0x70` |
| CreateObject | `0x36` | `0x36` | `0x36` |
| SetMultiVariables | `0x34` | `0x34` | `0x34` |
| GetVarSubStreamed | `0x36` | `0x34` | `0x34` |
| Explore | `0x34` | `0x34` | `0x34` |
| DeleteObject | `0x36` | `0x34` | `0x34` |

Once the new codes from the C# library are added, they can be added to the same frozenset to keep compatibility.

PS: I checked if this was the cause why https://github.com/gijzelaerr/python-snap7/pull/820 was not working, but that was not the issue.

### DIFF
--- a/s7commplus/async_client.py
+++ b/s7commplus/async_client.py
@@ -45,6 +45,7 @@ from .connection import (
     _set_s7_groups,
 )
 from .protocol import (
+    FLAGS_34_FUNCTION_CODES,
     READ_FUNCTION_CODES,
     S7COMMPLUS_LOCAL_TSAP,
     S7COMMPLUS_REMOTE_TSAP,
@@ -737,8 +738,8 @@ class S7CommPlusAsyncClient:
                 0x0000,
                 seq_num,
                 self._session_id,
-                # Transport flags: 0x34 for GetMultiVariables and Explore, 0x36 otherwise.
-                0x34 if function_code in (FunctionCode.GET_MULTI_VARIABLES, FunctionCode.EXPLORE) else 0x36,
+                # Transport flags: 0x34 for the function codes the reference sends with 0x34.
+                0x34 if function_code in FLAGS_34_FUNCTION_CODES else 0x36,
             )
 
             integrity_id_bytes = b""
@@ -992,7 +993,7 @@ class S7CommPlusAsyncClient:
             0x0000,
             seq_num,
             self._session_id,
-            0x36,
+            0x34,
         )
         request += struct.pack(">I", 0)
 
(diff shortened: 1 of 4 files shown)


gijzelaerr (changes requested):
The transport-flag mapping is protocol behavior and currently has no regression coverage in this PR. Please add byte-level tests for both sync and async request builders that demonstrate:

- every function in `FLAGS_34_FUNCTION_CODES` emits `0x34`;
- an ordinary `CreateObject` remains `0x36`;
- session-key requests retain their existing `0x34` override;
- `_delete_session()` emits `0x34` in both clients.

The hardware result is valuable, but without these tests a later request-builder refactor can silently restore the framing mismatch. This is also a prerequisite for #849, whose capture fixtures currently fail without this branch, so the boundary needs to be locked down before merge.

bvanelli:
I have pushed https://github.com/gijzelaerr/python-snap7/pull/848/changes/1946bd6c0a8137f5f01425a00b05cb520253ff1a to test regressions in the wire format from what you mentioned, but testing there is still a little bit awkward because the payloads are generated in-line, and also not shared between sync and async library. This refactor belongs in a different PR, so I included the basic testing with the `OSError` trick. This can be enhanced in the future with real traces.

gijzelaerr (approved):
The transport-flag mapping now has byte-level coverage for session CreateObject, SessionKey overrides, and sync/async DeleteObject, with the shared function-code set covering the common request builders. I ran the S7CommPlus framing and TLS suites locally: 93 passed.

PR #855, opened 2026-09-02
fix: narrow write_multi_vars item types

gijzelaerr:
## Summary

- replace the redundant list cast with per-item type narrowing
- reject mixed S7DataItem and dictionary lists with a clear TypeError
- keep compatibility with both the pinned mypy hook and newer mypy releases

## Validation

- uv run --frozen pre-commit run --all-files
- pytest -q (1821 passed, 78 skipped)
- mypy snap7/client.py
- uv build --no-sources

### DIFF
--- a/snap7/client.py
+++ b/snap7/client.py
@@ -1383,9 +1383,10 @@ class Client(ClientMixin):
             raise ValueError(f"Too many items: {len(items)} exceeds MAX_VARS ({self.MAX_VARS})")
 
         # Handle S7DataItem list (ctypes)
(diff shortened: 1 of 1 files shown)


gijzelaerr:
@bvanelli Given your recent S7CommPlus framing and multi-operation contributions, could you review this small write_multi_vars typing correction?

bvanelli (approved):
LGTM! Items still accepts dict[str, Any]. You could in theory make it stricter with a typedict. Right now, it still can fail with a KeyError, while the type hint is correct.

Issue #853, opened 2026-09-01
Server Event/Callback handling


NiklasReisser:
Hi, 

first of all, thanks a lot for this handy project!

I have an problem with the server implementation though:

The callback set in `set_events_callback` appears to never be called (I both tried it and looked at the code, for both Snap7 and S7Commplus)
Which means the callback appears to be useless. Happy to open a PR, however, I'm not really a subject matter expert, and 
would need some guidance as to what kind of events exist and should cause this to be triggered. 
I assume a Write from the client would certainly qualify. Would that be the case for all client requests? Anything else?

A similar question relies to the event queue related to `pick_event`, which seems to only ever receive a start event. However, that appears to be removed in the new S7CommPlus version.

Thanks and best regards

Niklas

gijzelaerr:
Confirmed and addressed in #856. The public snap7.Server API now queues events and invokes the general callback for server lifecycle, client connection, read, and write events; the dedicated read callback is also invoked, and callback failures cannot terminate the server loop. The event codes and read/write parameters follow the native Snap7 layout. The S7CommPlus emulator currently has only a private unused callback field and no public set_events_callback contract, so I have not invented a second incompatible callback API in this fix. @NiklasReisser, could you retest your snap7.Server workflow against #856, including a client write and pick_event()?

NiklasReisser:
Hi, looks good to me - left a few questions & comments on the PR. 

What's your plan for the next release?

gijzelaerr:
hey, human Gijs here. My plan was to have full s7comm+ v1,v2,v3 support for 4.0. we are getting close. What do you think?

NiklasReisser:
Hey, to be fully honest, I'm not really an expert on the matter. I am actually working on an application communicating with a S7 server, and I want to mock the server to test my application. So I don't know too much about s7comms. :-)

gijzelaerr:
Implemented in #856, now merged. The reporter confirmed the callback/event behavior looks good, so this is complete. Thanks for reporting and validating it.

PR #856, opened 2026-09-02
fix(server): dispatch callbacks and queue events

gijzelaerr:
## Summary

- dispatch native-compatible lifecycle, client, read, and write server events
- deliver events through both callbacks and the pick_event queue
- make queue access thread-safe and isolate callback failures from the server loop
- correct the startup event code to evcServerStarted

## Validation

- uv run --frozen pre-commit run --all-files
- pytest -q (1822 passed, 78 skipped)
- uv build --no-sources

Addresses #853.

### DIFF
--- a/snap7/server/__init__.py
+++ b/snap7/server/__init__.py
@@ -6,23 +6,35 @@ projects, use ``s7.Server`` instead, which supports both legacy S7 and
 S7CommPlus clients.
 """
 
+import logging
 import socket
 import struct
 import sys
 import threading
 import time
-import logging
-from typing import Dict, Optional, List, Callable, Any, Tuple, Type, Union
-from types import TracebackType
-from enum import IntEnum
 from ctypes import Array, c_char
+from enum import IntEnum
+from types import TracebackType
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
 
-from ..s7protocol import S7Protocol, S7Function, S7PDUType, S7UserDataGroup, S7UserDataSubfunction
 from ..datatypes import S7Area, S7WordLen
 from ..error import S7ConnectionError, S7ProtocolError
-from ..type import SrvArea, SrvEvent, Parameter
+from ..s7protocol import S7Function, S7PDUType, S7Protocol, S7UserDataGroup, S7UserDataSubfunction
+from ..type import Parameter, SrvArea, SrvEvent
 
 logger = logging.getLogger(__name__)
+
+
+# Event codes from the Snap7 server API.
+EVC_SERVER_STARTED = 0x00000001
+EVC_SERVER_STOPPED = 0x00000002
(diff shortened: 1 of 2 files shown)


tests/test_server.py
```diff
@@ -98,16 +99,71 @@ def test_unregister_area(self) -> None:
...
+        self.assertEqual(
+            [(event.EvtParam1, event.EvtParam2, event.EvtParam3, event.EvtParam4) for event in events],
+            [(SrvArea.DB, 1, 2, 3), (SrvArea.DB, 1, 4, 2)],
+        )
+        read_event = self.server.pick_event()
+        write_event = self.server.pick_event()
+        self.assertIsInstance(read_event, SrvEvent)
+        self.assertIsInstance(write_event, SrvEvent)
+        assert isinstance(read_event, SrvEvent)
+        assert isinstance(write_event, SrvEvent)
```
NiklasReisser, line 127: Why the duplicated check? Wouldn't it be more important to check the local events list filed by the callback?

snap7/server/__init__.py
```diff
@@ -97,6 +109,7 @@ def __init__(self, log: bool = True, max_clients: int = 64, **kwargs: object) ->
 
         # Event queue for pick_event
         self._event_queue: List[SrvEvent] = []
+        self._event_lock = threading.Lock()
```
NiklasReisser: Consider using https://docs.python.org/3/library/queue.html to prevent the list from filling up too much. Also, you wouldn't need the lock.

NiklasReisser (commented):
To be honest, I don't fully understand the way pick_event is supposed to be used. 
Should I busy wait in my application? 
Or rather call pick_event in my callback to prevent the queue from overflowing?

snap7/server/__init__.py
gijzelaerr: Addressed in 0de7891: the list and explicit lock are replaced with a thread-safe queue.Queue bounded at 1024 events. Protocol handling never blocks; when full, the oldest event is discarded so the queue retains the most recent bounded history. I also added coverage for the capacity and overflow behavior.

tests/test_server.py
gijzelaerr: Addressed in 0de7891: I removed the duplicated isinstance assertions and now explicitly verify that the events captured by the callback are the same events available through pick_event.

gijzelaerr:
@NiklasReisser Thanks — pick_event() is a non-blocking polling API, so it should not be called from inside the callback and callers do not need to busy-wait when they can use callbacks. The callback is the push-style API; pick_event() is the polling alternative. I documented that distinction, bounded the independent event queue, and added tests in 0de7891. Could you take another look?

NiklasReisser:
Looks good to me.  :+1:

PR #861, opened 2026-09-08
fix(s7commplus): make session setup atomic

### LINKED ISSUE #831: Make S7CommPlus session setup and SessionKey activation atomic
## Problem

The sync client installs `self._session_key` and enables integrity counters before the PLC accepts SetupSession. Both sync and async clients can then set `connected = True` even when session setup returns false.

This produces an invalid public state: transport TCP/COTP may be connected, but the S7CommPlus session is unusable or locally authenticated with a key the PLC rejected.

## Scope

- Keep generated key material pending until SetupSession succeeds.
- Commit the key and integrity state only after validating the matching successful response.
- Treat mandatory session-setup failure as a connection failure.
- Ensure cleanup clears pending and active authentication state.
- Define internal transport-connected versus protocol-session-ready states without changing unrelated public APIs.

## Acceptance criteria

- `connected` is never true after mandatory SetupSession rejection or malformed response.
- Rejected authentication does not leave `_session_key` or integrity tracking active.
- Successful legacy and TLS setup still transition to connected exactly once.
- Disconnect is safe from every intermediate state.
- Tests cover success, PLC error, malformed response and socket failure during setup.

## Out of scope

- HMAC verification itself
- Key renewal
- Async legacy-auth parity

gijzelaerr:
S7CommPlus could expose a connected client or activate generated SessionKey material before SetupSession succeeded. Keep key material pending, track internal readiness separately, and fail connection setup when ServerSessionVersion is missing or setup is rejected. Sync/async failures clear intermediate authentication state. This also addresses the false success indication reported in #877; its underlying CreateObject rejection still needs a failing Python capture.

Fail with an actionable installation error when legacy SessionKey authentication dependencies cannot load, instead of silently sending a different unauthenticated SetupSession.

The Python/TIA captures in #872 also reveal a legacy GetVarSubStreamed qualifier mismatch: the Python request has two extra bytes. Use the captured zero-valued qualifier/request layout and reject failed or malformed challenge replies without reusing the earlier CreateObject challenge. A packet regression matches the accepted TIA request bytes and verifies generated HMAC framing with a synthetic key. Hardware retesting is still needed; this does not claim the PLC connection issue is resolved.

Dependency #859 is merged. This PR now targets master and includes a forward merge of master; no pushed history was rewritten.

Validation after merging master: 2,037 passed, 78 skipped; source/wheel builds passed. Full mypy has the same 207 diagnostics as the updated baseline, with no new diagnostics. Regression coverage includes missing dependencies, failure-state cleanup, captured request layout, rejected challenge replies, and successful use of a fresh challenge.

Hardware evidence: russwing verified that missing cryptography prevents SetupSession from being sent and closes the ISO connection cleanly. Their subsequent report in #876 confirms successful connections using TLS after fixing group selection; their earlier plaintext reset is not evidence that the atomic-setup change fails on that CPU. Rejected/malformed-SetupSession guards still lack direct hardware coverage.

Fixes #831
Related: #872, #877, #876

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ Major release: new `s7commplus` package with S7CommPlus protocol support.
 * S7CommPlus symbolic data subscriptions and notification decoding (experimental)
 * TIA Portal XML import for SymbolTable (`SymbolTable.from_tia_xml()`) (experimental)
 * S7CommPlus CPU state reading and block transfer (upload/download)
+* Fix the legacy SecurityKey descriptor to identify the newly generated
+  session key instead of an all-zero placeholder.
 * **Symbolic (LID-based) access for optimized DBs** (experimental):
   `Tag.from_access_string("8A0E0001.A", "REAL")` creates a symbolic Tag;
   `client.read_tag(tag)` routes to S7CommPlus LID-based access via the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ CHANGES
 
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
+* Correct legacy GetVarSubStreamed qualifiers and reject unusable authentication challenges.
+
 * New `s7commplus` package for S7CommPlus protocol (S7-1200/1500)
 * S7CommPlus V1, V2 (TLS), and V3 support for S7-1200/1500
 * S7CommPlus area read/write (M, I, Q, counters, timers)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@ Major release: new `s7commplus` package with S7CommPlus protocol support.
 * S7CommPlus CPU state reading and block transfer (upload/download)
 * Fix the legacy SecurityKey descriptor to identify the newly generated
   session key instead of an all-zero placeholder.
+* Keep SessionKey activation and public connection state pending until the PLC
+  accepts session setup, with complete cleanup on rejection or transport error.
 * **Symbolic (LID-based) access for optimized DBs** (experimental):
   `Tag.from_access_string("8A0E0001.A", "REAL")` creates a symbolic Tag;
   `client.read_tag(tag)` routes to S7CommPlus LID-based access via the
--- a/doc/connecting.rst
+++ b/doc/connecting.rst
@@ -138,6 +138,11 @@ with the ``s7commplus`` extra:
 
    pip install 'python-snap7[s7commplus]'
 
+For a source checkout, use ``python -m pip install -e '.[s7commplus]'``.
+If the legacy SessionKey authentication path cannot import its dependencies,
+connection setup raises an error with these installation instructions and
+releases the connection; it does not continue with an unauthenticated setup.
+
 .. note::
 
    Older S7-1200 firmware (FW < 4.5) negotiates V1 of the S7CommPlus
--- a/s7commplus/async_client.py
+++ b/s7commplus/async_client.py
@@ -101,6 +101,8 @@ class S7CommPlusAsyncClient:
         self._subscription_container_id: int = 0
         self._sequence_number: int = 0
         self._protocol_version: int = 0
+        self._transport_connected = False
+        self._session_ready = False
         self._connected = False
         self._lock = asyncio.Lock()
 
@@ -188,6 +190,7 @@ class S7CommPlusAsyncClient:
 
         # TCP connect
         self._reader, self._writer = await asyncio.open_connection(host, port)
+        self._transport_connected = True
 
         try:
             # Step 1: COTP handshake with S7CommPlus TSAP values
@@ -208,7 +211,22 @@ class S7CommPlusAsyncClient:
             if self._tls_active:
                 self._protocol_version = ProtocolVersion.V2
 
-            # Step 5: Version-specific validation
+            # Step 5: Session setup. A transport and CreateObject response do
+            # not make the public client usable until the PLC accepts setup.
+            if self._server_session_version is None:
+                from snap7.error import S7ConnectionError
+
+                raise S7ConnectionError(
+                    "PLC did not provide a usable ServerSessionVersion attribute; S7CommPlus session setup cannot continue"
+                )
+            self._session_setup_ok = await self._setup_session()
+            if not self._session_setup_ok:
+                from snap7.error import S7ConnectionError
+
+                raise S7ConnectionError("S7CommPlus session setup was rejected by the PLC")
+            self._session_ready = True
+
+            # Step 6: Version-specific validation
             if self._protocol_version >= ProtocolVersion.V3:
                 if not use_tls:
                     logger.warning(
@@ -224,20 +242,11 @@ class S7CommPlusAsyncClient:
                 self._integrity_id_write = 0
                 logger.info("V2 IntegrityId tracking enabled")
 
+            self._protection_level = await self._get_effective_protection_level()
+            if self._protection_level is not None:
+                logger.info(f"PLC reports protection level: {self._protection_level}")
+
             self._connected = True
-
-            # Step 6: Session setup - echo ServerSessionVersion back to PLC
-            if self._server_session_version is not None:
-                self._session_setup_ok = await self._setup_session()
-            else:
-                logger.warning("PLC did not provide ServerSessionVersion - session setup incomplete")
-                self._session_setup_ok = False
-
-            # Only a session that completed setup answers attribute reads.
-            if self._session_setup_ok:
-                self._protection_level = await self._get_effective_protection_level()
-                if self._protection_level is not None:
-                    logger.info(f"PLC reports protection level: {self._protection_level}")
 
             logger.info(
                 f"Async S7CommPlus connected to {host}:{port}, "
@@ -452,13 +461,15 @@ class S7CommPlusAsyncClient:
 
     async def disconnect(self) -> None:
         """Disconnect from PLC."""
-        if self._connected and self._session_id:
+        if self._session_ready and self._session_id:
             try:
                 await self._delete_session()
             except Exception:
                 pass
 
         self._connected = False
+        self._session_ready = False
+        self._transport_connected = False
         self._session_id = 0
         self._subscription_container_id = 0
         self._sequence_number = 0
@@ -838,7 +849,7 @@ class S7CommPlusAsyncClient:
             Response payload (after the 10-byte response header).
         """
         async with self._lock:
-            if not self._connected or self._writer is None or self._reader is None:
+            if not (self._connected or self._transport_connected) or self._writer is None or self._reader is None:
                 raise RuntimeError("Not connected")
 
             seq_num = self._next_sequence_number()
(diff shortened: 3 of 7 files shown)


gijzelaerr:
@russwing @EliTheCoder could you each help with one targeted hardware check on this branch?

- @russwing: your plaintext S7-1500 CPU 1511F-1 PN (FW V2.9) reached the SetupSession failure path in #710, so it would give us an independent V1/S7-1500 check of the new atomic failure handling.
- @EliTheCoder: your 6ES7 515-2TM01-0AB0 (FW V2.9) is a known-clean V2/TLS target, so a connect plus browse/read smoke test would check that this does not regress the successful TLS path.

This PR is stacked on #859, so testing this head covers the generated SessionKey descriptor fix there as well. A connect attempt and confirmation that failed setup leaves the client disconnected/clean would be sufficient; no sensitive captures or project data needed.

russwing:
### Hardware check — plaintext S7-1500 CPU 1511F-1 PN, V2.9

@gijzelaerr — ran it. **No regression, all cleanliness checks pass. But the new atomic-failure branches are not reached on this CPU**, so this is a weaker signal than you were hoping for. The reason is the useful part.

**Target** — confirmed live via classic S7comm at the same moment, so the box is healthy and the reset below is specific to S7CommPlus:

```
module      CPU 1511F-1 PN
order code  6ES7 511-1FK02-0AB0
cpu_state   S7CpuStatusRun
```

**Tested** `f29551ff992c` (this PR head, so #859's `883c9ba` is included), `pip install -e ".[s7commplus]"`, Python 3.11.2, Linux, plaintext :102.

### What happens

The `SetupSession` frame goes out — 402 bytes on the wire, including the 180-byte generated SessionKey blob — and the PLC **resets the TCP connection instead of replying**:

```
=== CreateObject === sending (192 bytes)
CreateObject returned non-zero 0x4000800000000011; continuing to parse the returned session data
ServerSessionVersion captured (87 bytes)
Public key fingerprint captured: 00:181B7B0847D11694
Session challenge captured (20 bytes)
SessionKey auth blob generated (180 bytes)
=== SetupSession === sending (395 bytes)
→ ConnectionResetError: [Errno 104] Connection reset by peer
```

So it raises out of the transport, one level below your new code:

```
connection.py:505   self._session_setup_ok = self._setup_session()
connection.py:1375    response_frame = self._recv_s7_data()
snap7/connection.py:476  raise S7ConnectionError(f"Receive error: {e}")
```

**Neither new `raise` is reached.** `ServerSessionVersion` is present (so the first guard passes) and `_setup_session()` never returns `False` (so the second is never evaluated) — it dies inside the receive. This CPU therefore does not cover the rejection or malformed-response paths; your unit tests remain the only coverage for those.

### Delta vs `master` (`383852f3eb64`)

Identical on every observable, on the same box minutes apart. The only difference is that `_session_ready` exists and is correctly `False`; on master the attribute is absent.

That is expected given the above — a transport reset already propagated as an exception before this PR, so there was never a half-open client to fix *on this failure mode*.

### What this does check

The cleanup half — *"propagate async setup transport/protocol failures instead of swallowing them"* and *"clear pending and active authentication state during every failure cleanup"*. All hold, twice in a row:

| check | result |
|---|---|
| `connect()` does not silently half-succeed | pass |
| client left disconnected (`connected is False`) | pass |
| internal `_connected` / `_session_ready` / `_session_setup_ok` all `False` | pass |
| socket released (`_iso_conn.socket is None`, `_iso_conn.connected is False`) | pass |
| no fd leaked (fd count before/after, 2 attempts) | pass — delta 0 |
| `disconnect()` safe after a failed connect | pass |
| `disconnect()` idempotent | pass |
| second attempt identical to first, no residue | pass |
| `browse()` after failed connect → clean `Not connected` | pass (same on master) |

### On #859

With `cryptography` present the SessionKey blob is generated and encoded into the SetupSession frame (180 bytes, visible in the trace above), so the generated-descriptor path **is** exercised on real hardware. It does not change the outcome on this CPU — the PLC still resets — but it does not break the frame either.

### One packaging note

My first run used a plain `pip install -e .` and got:

```
WARNING s7commplus.connection: SessionKey auth failed: No module named 'cryptography'
```

It then continued and failed at the same place, so the surface symptom was identical and I nearly reported it as the same result. The `cryptography` dependency sits behind the `s7commplus` extra, and missing it silently degrades the auth path to a warning. Might be worth raising it to an error, or naming the extra in the connect-time message — a user who follows the plain install gets a quietly different code path.

### Offer

The reset itself is presumably the more interesting question, and it is the same one from #710. If a paired capture would help — tcpdump either side of the SetupSession, or the PLC's diagnostic buffer entry for the rejected connection — say the word and I will get it. On CPU config, with the caveat properly attached: **PUT/GET is enabled** — re-confirmed tonight, since classic S7comm read CPU info and an S7-1500 will not serve that otherwise. The **access level was Full access / no protection** when our PLC engineer set this box up on 2026-08-30, but I have not re-verified it since and cannot do so remotely (an S7-1500 does not expose the protection SZL — `get_protection()` returns `Object does not exist (0x0a)`). He is checking it in TIA before we make any further claim on it. So treat "not a permission refusal" as likely rather than established.



---

*Edited to correct two things in the evidence above, both found by re-checking my own probe rather than by anything new from the PLC:*

*1. The socket row originally read `_socket is None`. **That attribute does not exist on the connection object**, so `getattr` returned `None` and the row passed without testing anything. Re-measured against the attribute that does exist: after the failed connect, `_iso_conn.socket is None` and `_iso_conn.connected is False`. The claim holds — but it was a vacuous pass as originally written, and the fd-delta row was the only thing genuinely proving no leak.*

*2. The CPU-configuration statement was flatter than the evidence supports; it now carries its provenance and its age.*

russwing:
One extra data point from the same session that narrows this, which I should have put in the first comment.

I ran the check twice — once **without** `cryptography` installed and once with. Because `include_security_key = auth_result is not None`, those are two structurally different SetupSession frames:

| run | `auth_result` | frame | result |
|---|---|---|---|
| no `cryptography` | `None` | **122 bytes** — `ItemCount 1`, ServerSessionVersion only | RST at the same point |
| with `cryptography` | 180-byte blob | **395 bytes** — `ItemCount 2`, SecurityKey + ServerSessionVersion | RST at the same point |

Measured at the same offset in both frames, after `70 00 0c 83 34 70 00 0c 83`:

```
no cryptography    ... 70 00 0c 83  01 01  82 32 01 00 17 ...   <- ItemCount 1, AddressCount 1
with cryptography  ... 70 00 0c 83  02 02  8e 26 82 32 01 ...   <- ItemCount 2, AddressCount 2
```

**Identical outcome either way.** So on this CPU the SecurityKey item is not the variable — the reset does not depend on whether it is present or how it is encoded, which suggests #859's descriptor fix is not what this particular PLC is objecting to.

`_strip_paom_string_in_session_version` is applied in both, so the PAOM workaround is also not it.

gijzelaerr:
Thanks @russwing — the repeated cleanup checks are useful, and the corrected socket check makes the result much clearer. I will record this as a plaintext transport-reset cleanup/non-regression test. It does not exercise the new missing-ServerSessionVersion or rejected-SetupSession guards, and it does not establish that the generated SecurityKey descriptor is accepted by this CPU.

The reset with and without SecurityKey narrows the investigation, but it does not by itself rule out malformed common framing, multiple rejection causes, or the PAOM transformation shared by both attempts.

Yes, please provide the offered capture covering CreateObject through SetupSession/reset, the corresponding PLC diagnostic-buffer entry, and the access/protection settings once the PLC engineer has rechecked them. If a known-working S7CommPlus client is available against the same unchanged CPU, a comparable handshake capture would be especially useful. Please omit credentials and unrelated project data. We can keep the connection-reset investigation linked to #710 without treating it as a regression introduced here.

I have also prepared a dependency-handling change for this branch: an authentication import failure raises an actionable error naming the s7commplus extra and source-checkout install command, rather than continuing with a different SetupSession. A fresh-interpreter regression test blocks cryptography and verifies that setup is not sent, the socket is closed, and a second attempt leaves no authentication state. This change is currently local, pending branch publication.

@EliTheCoder the outstanding complementary check is still a successful TLS connect plus browse/read on your V2 target. There is no need to repeat the plaintext failure test on that CPU. #859 remains the base dependency; this PR should be retargeted after it lands.

gijzelaerr:
The dependency-handling fix is now published in 637f8bd on this PR. The description includes the current validation and the limits of the hardware coverage.

To test this exact revision with the required dependencies:

```sh
python -m pip install 'python-snap7[s7commplus] @ git+https://github.com/gijzelaerr/python-snap7.git@637f8bdc31bb444503ec3a51115fba4903f1169b'
```

With the extra installed, the existing plaintext SetupSession reset may still occur; this change does not claim to fix that reset. Without the authentication dependency, the legacy SessionKey path now stops with installation instructions before sending SetupSession.

The requested capture/diagnostic evidence and successful V2/TLS connect+browse/read remain the useful hardware checks. Local validation: 1,916 passed, 78 skipped; required hooks and source/wheel builds pass.

russwing:
### 637f8bd verified — plaintext S7-1500, CPU 1511F-1 PN, FW V2.9

Checked the dependency-handling change on the same unchanged CPU. Measured **on the wire**, not from the exception, since "setup is not sent" is a wire claim.

Three runs, same box, minutes apart. Lengths are TPKT frame bytes — app-layer payload plus 7 for TPKT + COTP DT, so the 402 below is the 395 the log prints.

| run | commit | `cryptography` | 4th client→PLC frame | outcome |
|---|---|---|---|---|
| A | `f29551ff` (previous head) | absent | **129 B** — COTP DT, S7CommPlus SetupSession | PLC RST |
| B | **`637f8bd`** | absent | **12 B** — COTP **Disconnect Request** | clean close; no S7CommPlus PDU sent |
| C | **`637f8bd`** | present | **402 B** — COTP DT, SetupSession | PLC RST, unchanged |

Run B's fourth frame in full:

```
03 00 00 0c   TPKT, length 12
06 80         COTP LI = 6, PDU type 0x80 = Disconnect Request
00 01 00 01 00
```

followed by a TCP FIN from the client. The RST after it is the PLC answering our FIN, not a rejection.

So the change does what the description says on this CPU: **the SetupSession that the previous head sent without an auth blob is no longer sent at all**, and the ISO connection is released cleanly. The error is raised rather than logged, and it names both the extra and the source-checkout command.

Run C is the non-regression check — with the extra installed the frame and the reset are unchanged from the previous head.

### Two observations from the same captures

**CreateObject is not minimal on this branch.** The 199-byte CreateObject carries nine attributes and a `SubscriptionContainer`:

```
0x00d0:  5375 6273 6372 6970 7469 6f6e 436f 6e74   SubscriptionCont
0x00e0:  6169 6e65 72                              ainer
```

Worth noting against the older reading of #710, where the client was described as introducing itself too minimally versus TIA. That is no longer what the frame looks like, and the CPU still returns `0x4000800000000011`. This does not rule anything out — it just removes one candidate explanation.

**`protection_level()` is not reachable here.** I tried it to answer your question about the access/protection settings without waiting on our engineer. It is only callable once `connect()` has succeeded, so it raises along with the session. Those settings still need TIA at our end; there is no remote path to them.

### On the capture

Ready — the pcap covers CreateObject through SetupSession and the reset. I would rather send it alongside the TIA capture of the same exchange against the same unchanged CPU, since the pair is what makes it a diff instead of another look at our own failure. Our PLC engineer has that, the diagnostic-buffer entry and the protection settings this week. If our half is useful to you before then, say so and I will post it now.

gijzelaerr:
Thanks @russwing — the wire comparison verifies the dependency fix much more directly. On 637f8bd without cryptography, the fourth frame is a COTP Disconnect Request followed by closure, with no SetupSession sent; with the extra installed, the existing SetupSession/reset behavior remains unchanged. I am updating the PR's hardware-validation note accordingly. The rejected/malformed-SetupSession guards and a successful TLS path remain separate coverage requirements.

Yes, please post the Python-side capture now. It will let us inspect the failing handshake while the same-CPU TIA capture, diagnostic-buffer entry and confirmed protection settings follow. Keep the paired capture against the same unchanged CPU when it is available.

Agreed on CreateObject: the old "minimal introduction" explanation is not established for this branch, which already carries multiple attributes and a SubscriptionContainer. Also, `protection_level` is a property populated after successful setup, so it cannot replace the TIA settings check on this failing connection.

There is now a separate TIA capture from a CPU 1511C FW 2.1.0 in #872. It starts directly with CreateObject, receives successful CreateObject and SetupSession replies, then continues with V3 traffic. That gives us another comparison target, but it is a different CPU/firmware and does not establish the cause of your reset.

PR #871, opened 2026-09-08
feat(s7commplus): add batched symbolic read

bvanelli:
Introduces the `read_symbolic_multi`, which is a generalization of the `read_symbolic`. It is not only backwards compatible, but allows the user to read an entire range of variables with a single request. Tested against a real S7 1500, FW 2.9.8, using a batch of 20 variables.

The failure mode when you provide a rejected variable is to log the the hex error, handled by the `_parse_read_response`:

```
ERROR s7commplus.client: _parse_read_response: PLC returned error 0x20278000009A0013
```

Since the error is swallowed by the parser, there is a length check for the response, and if variables are missing, a `Runtime` is raised. This should only guard the occasion where the return is `[]`, but the client expected variables.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ CHANGES
 
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
+* Validate batched symbolic read item coverage and preserve explicit PLC errors.
+
 * New `s7commplus` package for S7CommPlus protocol (S7-1200/1500)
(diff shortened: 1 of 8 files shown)


gijzelaerr (approved):
Reviewed the updated merge head aa7f661. The maintainer response-validation commit is present, the merge resolution is clean, the pinned pre-commit suite passes, source and wheel builds pass, and the full local suite passes (1,991 passed, 82 skipped).

PR #879, opened 2026-09-10
fix(s7commplus): decode CPU operating state

### LINKED ISSUE #878: s7commplus: get_cpu_state() returns "RUN" for a stopped CPU — the documented "STOP" return is unreachable
Found while validating the S7CommPlus TLS fix on real hardware (#710). @gijzelaerr asked for this as a focused issue.

## The defect is provable from the source alone

Both `client.py` and `async_client.py`:

```python
payload = _build_explore_request(Ids.NATIVE_THE_CPU_EXEC_UNIT_RID, [])
response = self._connection.send_request(FunctionCode.EXPLORE, payload, integrity_tail=5, reassemble=True)
# Parse for operating state attribute — return "RUN" as default
# since a responding PLC is typically running
return "RUN" if response else "UNKNOWN"
```

**The documented `"STOP"` return value is unreachable.** The docstring promises `"RUN"`, `"STOP"` or `"UNKNOWN"`; the implementation can only produce two of them. No CPU state can ever yield `"STOP"`.

The premise in the comment — *"a responding PLC is typically running"* — is the part that does not hold. **A stopped S7 CPU keeps answering.** Siemens documents this for the OPC UA server (*"An activated OPC UA server remains in operation even if the CPU switches to STOP ... you will get what were the latest values before the CPU switched to STOP"*, S7-1500 communication function manual §11.3.1.4), and we measured the same for S7CommPlus.

## Measured on hardware

Two CPU families, two firmware generations, two protection levels. Each STOP was performed and confirmed by our automation engineer, not inferred from an instrument:

| CPU | firmware | protection | state | `get_cpu_state()` |
|---|---|---|---|---|
| `6ES7 212-1BE40-0XB0` (1212C) | V4.7.3 | 1 | **STOP**, held ~7.5 min | `RUN` |
| `6ES7 212-1BE40-0XB0` (1212C) | V4.7.3 | 1 | RUN | `RUN` |
| `6ES7 511-1FK02-0AB0` (1511F-1 PN) | V2.9.7 | 0 | **STOP** | `RUN` |
| `6ES7 511-1FK02-0AB0` (1511F-1 PN) | V2.9.7 | 0 | RUN | `RUN` |

Sampled continuously across a full RUN -> STOP -> RUN transition on the 1212C: **153 samples, never anything but `RUN`.** For comparison, the same CPU's web-server start page reported `STOP` correctly throughout.

## Raw EXPLORE response — RUN

The response is 701 bytes and clearly structured, so there is an attribute set to decode rather than a flag to find:

```
# host=10.99.100.50 label=RUN
# protocol_version=2 protection_level=1 tls_active=True
# get_cpu_state() -> 'RUN'
# EXPLORE(NATIVE_THE_CPU_EXEC_UNIT_RID) response: 701 bytes
0000  00 00 00 00 30 07 a1 00 00 00 34 91 03 30 92 59  ....0.....4..0.Y
0010  06 a3 81 69 00 15 0d 43 50 55 20 65 78 65 63 20  ...i...CPU exec 
0020  75 6e 69 74 a3 93 15 00 05 00 a3 93 16 00 04 00  unit............
0030  a3 93 13 00 04 00 a3 91 3d 00 17 00 00 0d 99 9b  ........=.......
0040  1a 00 08 00 9b 1b 00 0c 00 00 00 00 9b 1c 00 0b  ................
0050  00 01 9b 1d 00 0c 00 00 80 00 9b 1e 00 08 08 9b  ................
0060  1f 00 14 00 00 00 a3 91 4c 00 08 00 a3 91 52 00  ........L.....R.
0070  01 00 a3 91 53 00 08 05 a3 91 54 00 03 00 00 a3  ....S.....T.....
0080  91 55 00 02 00 a3 91 56 00 02 08 a3 91 57 00 02  .U.....V.....W..
0090  02 a3 91 58 00 04 00 a3 91 59 00 14 00 04 00 01  ...X.....Y......
00a0  00 22 a3 91 5a 00 03 00 34 a3 91 5b 00 03 00 00  ."..Z...4..[....
00b0  a3 91 5c 00 0a 00 a3 9c 67 00 0b 00 00 a3 9c 68  ..\.....g......h
00c0  00 08 00 a3 9f 2e 00 17 00 00 0f d7 9f 58 00 03  .............X..
00d0  00 00 9f 59 00 03 00 00 9f 5a 00 03 00 00 9f 5b  ...Y.....Z.....[
00e0  00 03 00 01 9f 5c 00 04 00 9f 5d 00 03 00 06 00  .....\....].....
00f0  a3 a4 19 00 0b 00 01 a3 a4 1a 00 08 7f a3 a4 1b  ................
...
0270  22 31 34 2e 30 2e 30 2e 30 22 23 32 37 3d 22 41  "14.0.0.0"#27="A
0280  51 41 41 41 4a 63 41 41 41 41 41 41 41 41 41 43  QAAAJcAAAAAAAAAC
0290  6b 78 76 59 32 46 73 66 6b 56 34 5a 57 4d 43 4e  kxvY2FsfkV4ZWMCN
02a0  54 49 3d 22 23 35 37 3d 22 31 22 a3 a4 52 40 14  TI="#57="1"..R@.
02b0  00 a4 91 5d 00 00 00 31 a2 00 00 00 00           ...]...1.....
```

## The STOP capture

I have deliberately not posted a STOP capture yet. Stopping a controller needs our automation engineer, and the useful artefact is **the same unchanged CPU in both states so the diff isolates the attribute** — a single-state dump invites another guess, which is what you asked to avoid. It will follow as a comment here shortly.

If a particular capture form would help — a fuller `EXPLORE` of a different RID, or a scrubbed pcap of the exchange — say which and I will take that instead while the CPU is stopped, since the window is short and I would rather capture the right thing once.

gijzelaerr:
## Summary

- decode the two corroborating CPU execution attributes identified by the RUN/STOP hardware capture in #878
- report RUN only for executing=1 plus operating-mode=7, and STOP only for executing=0 plus operating-mode=0
- return UNKNOWN when either attribute is absent, malformed, contradictory, duplicated inconsistently, or carries an unrecognized value
- share the parser between synchronous and asynchronous clients

The conservative two-attribute requirement avoids interpreting an absent/default-initialized field as STOP.

## Validation

- `uv run --frozen --extra test --extra s7commplus pytest -q` (2040 passed, 82 skipped)
- `uv build --no-sources`
- capture-derived RUN/STOP attribute sequences plus malformed, partial, contradictory, and duplicate cases

## Hardware validation requested

@russwing Could you test this exact branch against the same S7-1200 and S7-1500 RUN → STOP → RUN setup? In particular, please confirm that `get_cpu_state()` follows all three phases on both clients/CPUs and never reports STOP when the state attributes are absent.

Fixes #878

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ CHANGES
 
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
+* Decode corroborating CPU execution attributes so S7CommPlus `get_cpu_state()`
+  distinguishes RUN from STOP and returns UNKNOWN for absent or inconsistent state.
 * Support LOGO reconnection/heartbeat options and preserve explicit TSAPs.
 * Correct legacy GetVarSubStreamed qualifiers and reject unusable authentication challenges.
 * Validate batched symbolic read item coverage and preserve explicit PLC errors.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,8 @@ CHANGES
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
 * Decode corroborating CPU execution attributes so S7CommPlus `get_cpu_state()`
-  distinguishes RUN from STOP and returns UNKNOWN for absent or inconsistent state.
+  distinguishes RUN from STOP on S7-1500 and returns UNKNOWN for absent or
+  inconsistent state attributes, including S7-1200 responses that omit them.
 * Support LOGO reconnection/heartbeat options and preserve explicit TSAPs.
 * Correct legacy GetVarSubStreamed qualifiers and reject unusable authentication challenges.
 * Validate batched symbolic read item coverage and preserve explicit PLC errors.
--- a/s7commplus/async_client.py
+++ b/s7commplus/async_client.py
@@ -27,6 +27,7 @@ from .client import (
     _build_symbolic_write_payload,
     _build_write_payload,
     _parse_explore_datablocks,
+    _parse_cpu_state,
     _parse_read_response,
     _parse_write_response,
 )
@@ -586,7 +587,7 @@ class S7CommPlusAsyncClient:
         """
         payload = _build_explore_request(Ids.NATIVE_THE_CPU_EXEC_UNIT_RID, [])
         response = await self._send_request(FunctionCode.EXPLORE, payload, integrity_tail=5, reassemble=True)
-        return "RUN" if response else "UNKNOWN"
+        return _parse_cpu_state(response)
 
     async def upload_block(self, block_type: int, block_number: int) -> bytes:
         """Upload (read) a program block from the PLC.
--- a/s7commplus/client.py
+++ b/s7commplus/client.py
@@ -528,9 +528,7 @@ class S7CommPlusClient:
         # Read the CPU exec unit object to get the running state
         payload = _build_explore_request(Ids.NATIVE_THE_CPU_EXEC_UNIT_RID, [])
         response = self._connection.send_request(FunctionCode.EXPLORE, payload, integrity_tail=5, reassemble=True)
-        # Parse for operating state attribute — return "RUN" as default
-        # since a responding PLC is typically running
-        return "RUN" if response else "UNKNOWN"
+        return _parse_cpu_state(response)
 
     def upload_block(self, block_type: int, block_number: int) -> bytes:
         """Upload (read) a program block from the PLC.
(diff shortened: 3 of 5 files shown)


russwing:
Tested. **Correct on the S7-1500. On the S7-1200 it returns `UNKNOWN` in RUN — and that is the honest answer, not a parser bug.**

## S7-1500 — `6ES7 511-1FK02-0AB0` (1511F-1 PN), FW V2.9.7

I replayed this branch's `_parse_cpu_state` over **109 EXPLORE responses captured across the RUN → STOP → RUN transition**, each labelled independently by the CPU's own web server (polled every 2 s alongside the S7CommPlus sampling):

```
RUN     60/ 60 correct
STOP    49/ 49 correct
```

Live on the same CPU, currently in RUN — **sync `'RUN'`, async `'RUN'`**. The STOP and return-to-RUN phases are from the capture replay rather than a second live transition, since stopping the controller needs our automation engineer at the panel.

Every degenerate case returns `UNKNOWN`, never `STOP`:

| case | result |
|---|---|
| empty response | `UNKNOWN` |
| truncated to 32 bytes | `UNKNOWN` |
| attribute markers stripped | `UNKNOWN` |
| all-zero bytes (default-initialised) | `UNKNOWN` |
| `executing=1` with `mode=0` (contradictory) | `UNKNOWN` |
| `executing=0` with `mode=7` (contradictory) | `UNKNOWN` |
| unrecognised mode value (9) | `UNKNOWN` |

The two-attribute requirement does exactly what you designed it for.

## S7-1200 — `6ES7 212-1BE40-0XB0` (1212C), FW V4.7.3

Live, in RUN: **sync `'UNKNOWN'`, async `'UNKNOWN'`.**

Not a decode failure — **the attributes are not there.** Enumerating every attribute marker in the 1212C's 701-byte `EXPLORE(NATIVE_THE_CPU_EXEC_UNIT_RID)` reply:

```
attributes in 0x1F00-0x2000: none        (1212C, 701 bytes)
attributes in 0x1F00-0x2000: 0x1f54 0x1f80 0x1f81 0x1fe5   (1511F, 651 bytes)
```

The 1212C's attribute IDs on this object top out at `0x001d8e`. `0x1F80`/`0x1F81` appear to be S7-1500-only.

**So the three-phase test cannot be satisfied on the S7-1200 by any amount of hardware time** — if the CPU cannot report RUN it cannot report STOP either. Worth knowing before you wait on more captures from me.

## The judgement call this leaves you

On S7-1200 the behaviour changes from **always `"RUN"`** (wrong, but a value people may be branching on) to **always `"UNKNOWN"`** (correct). I think `UNKNOWN` is right and should be documented rather than papered over — an absent attribute is exactly the case your two-attribute rule exists to refuse, and a CPU that cannot answer should say so. But it is a visible change for every S7-1200 user, so it probably belongs in CHANGES.md explicitly rather than as a side effect.

## Offer

If you want S7-1200 coverage, the next step is finding whether some other RID carries execution state there. I can run a RUN/STOP diff against a different object on the 1212C — that needs one more stop window from our engineer, so say which RID and I will get it in one pass rather than several.

Scripts and the raw labelled captures are to hand if any of it would be useful in your test fixtures.

gijzelaerr:
Updated this branch onto current `master` and clarified the documented hardware scope. The hardware report confirms 109/109 captured S7-1500 transitions decoded correctly (60 RUN, 49 STOP), with sync and async live reads agreeing. S7-1200 omits these attributes, so both clients now correctly return UNKNOWN rather than an unconditional RUN. Local validation on head `07d8deb`: all pre-commit hooks passed, 2071 tests passed (82 skipped), and source/wheel builds passed.

PR #882, opened 2026-09-11
fix(s7commplus): enforce async auth boundary

gijzelaerr:
## Summary

- parse async CreateObject authentication attributes with the shared parser
- reject legacy V1 SessionKey authentication before session setup with an actionable error
- document the synchronous/async V1-V3 authentication support matrix
- cover SessionKey detection and connect cleanup with mirrored async tests

## Validation

- `uv run pytest -q`: 2125 passed, 78 skipped
- `uv run pre-commit run --all-files`: passed
- `uv build --no-sources`: passed

Fixes gijzelaerr/s7commplus#5

PR #883, opened 2026-09-11
fix(s7commplus): correlate request responses

### LINKED ISSUE #833: Correlate S7CommPlus responses and route unsolicited frames separately
## Problem

`send_request()` decodes the response opcode, function code and sequence number only for logging. It does not verify that the frame matches the outstanding request.

Notifications, SystemEvents, delayed responses or otherwise unexpected frames can consequently be consumed as an ordinary request response. This also blocks a correct subscription API. SystemEvent payload parsing has a related report in #827.

## Scope

- Require response opcode/function/sequence to match the outstanding request.
- Recognize and route Notification and SystemEvent frames without treating them as responses.
- Reject unexpected or duplicate response frames with a protocol-specific error.
- Serialize sync requests per connection or otherwise guarantee unambiguous correlation.
- Use the same dispatcher model as the foundation for sync and async clients.

## Acceptance criteria

- A mismatched sequence or function never satisfies a request.
- A notification arriving before the response is preserved and dispatched.
- A SystemEvent produces the behavior decided in #827 rather than being parsed as data.
- Concurrent sync calls cannot interleave wire requests unsafely.
- Tests cover notification-before-response, wrong sequence, wrong function, duplicate response and connection close.

## Out of scope

- Public subscription lifecycle API
- Alarm decoding (#794)

gijzelaerr:
## Summary

- validate S7CommPlus response opcode, function, and sequence for sync and async requests
- discard a bounded number of delayed replies from older 16-bit sequences while continuing to wait for the current response
- retain protocol-defined generic ERROR replies while rejecting unrelated or future response functions/sequences
- route interleaved notifications into per-connection queues and consume SystemEvents separately
- serialize synchronous request/response exchanges to prevent wire interleaving
- report closed connections as connection errors

## Real-hardware evidence

The S7-1511C trace in #872 showed a DB-read request at sequence 7 consuming the delayed protection-level reply for sequence 6. The added regression covers that ordering for both sync and async clients. PR #881 is now stacked on this branch so the combined behavior can be tested on that PLC.

## Validation

- `uv run --frozen pytest -q`: 2087 passed, 82 skipped
- `uv run --frozen pre-commit run --all-files`: passed
- `uv build --no-sources`: passed

Fixes #833

### DIFF
--- a/s7commplus/connection.py
+++ b/s7commplus/connection.py
@@ -45,6 +45,7 @@ import os
 import ssl
 import struct
 import tempfile
+import threading
 from collections import deque
 from types import TracebackType
 from typing import Any, Optional, Type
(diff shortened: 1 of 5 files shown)


gijzelaerr:
@bvanelli Could you review the S7CommPlus protocol/security changes as one batch? The independent roots are #882, this PR, #887, and #885; #884 and #888 form the stacked continuation of this PR. Your earlier protocol traces and authentication reviews are especially relevant to the response correlation, integrity-ID/HMAC validation, and session-key fallback/renewal paths. Reviewing the roots first is enough; we can update the stacked branches after any findings.

s7commplus/connection.py
```diff
@@ -80,6 +81,71 @@
...
+        hash_length = data[0]
+        if len(data) <= 1 + hash_length:
+            raise S7ProtocolError("Truncated S7CommPlus V3 integrity envelope")
+        data = data[1 + hash_length :]
+    if not data:
+        raise S7ProtocolError("S7CommPlus frame has no application opcode")
+    return data[0]
+
+
+def _incoming_response_sequence(frame: bytes) -> int:
```
bvanelli: The validation branching of this and `_incoming_frame_opcode` is almost the same. I wonder if there are more cases and you can generalise it.

bvanelli (approved):
LGTM, i checked some behavioural properties:

- The `stale_responses` resets, so an otherwise healthy subscription cannot easility be killed.
- Code repetition in `_incoming_frame_opcode` and `_incoming_response_sequence` goes through different branches, not worth generalising.
- Lock now guarantees ordering for `send_request`, and there is no deadlock possibility due to the timeout.

PR #884, opened 2026-09-11
fix(s7commplus): verify authenticated responses

### LINKED ISSUE #832: Verify S7CommPlus V3 and legacy response HMACs
## Problem

For authenticated traffic, the client calculates HMAC-SHA256 for outgoing requests. Incoming responses merely have their hash-length byte and digest stripped and logged; the digest is never verified.

A corrupted, incorrectly keyed or tampered response is therefore accepted as valid application data. Fragment reassembly and Explore continuation handling also strip authenticated prefixes without verification.

## Scope

- Verify response digests with constant-time comparison before parsing application data.
- Cover single-frame responses and all fragment/reassembly paths.
- Implement the legacy first-fragment/cumulative digest semantics used by compatible implementations.
- Raise a dedicated, actionable S7CommPlus integrity error on mismatch.
- Add deterministic valid, wrong-key, modified-payload, truncated-digest and fragmented-response tests.

Server-side generation is related to #803, but client verification should be independently testable with captured/golden frames.

## Acceptance criteria

- No authenticated response reaches a payload parser before successful digest verification.
- Any changed digest-covered byte causes a deterministic exception and invalidates the connection.
- Invalid digest length and truncated frames are rejected.
- Valid multi-fragment responses still reassemble correctly.
- Comparison uses `hmac.compare_digest` or an equivalent constant-time primitive.

gijzelaerr:
Closes gijzelaerr/s7commplus#2.

## Summary
- verify every SessionKey-protected V3 response with `hmac.compare_digest` before opcode routing or payload parsing
- preserve the legacy cumulative HMAC state across fragmented responses
- reject invalid/truncated digest envelopes, unsigned downgrade frames, and fragment-version changes
- raise a dedicated `S7IntegrityError` and close/reset the connection without sending on the untrusted stream
- verify interleaved notifications before queueing and require the verified reassembly path for authenticated Explore continuations

This PR is intentionally stacked on gijzelaerr/python-snap7#883 because safe pre-parse verification depends on its response/notification dispatcher. Once gijzelaerr/python-snap7#883 merges, this PR can be retargeted to `master`.

## Verification
- `uv run pre-commit run --all-files`
- `.venv/bin/pytest` — 2,099 passed, 82 skipped
- `uv build` — wheel and sdist built successfully

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ Major release: new `s7commplus` package with S7CommPlus protocol support.
 
 * Correlate S7CommPlus responses by opcode, function, and sequence; preserve
   interleaved notifications and serialize synchronous wire requests.
+* Verify authenticated V3 responses and cumulative fragment digests before
+  parsing, invalidating the connection with a dedicated integrity error on any
+  mismatch or truncated envelope.
 * Decode corroborating CPU execution attributes so S7CommPlus `get_cpu_state()`
   distinguishes RUN from STOP on S7-1500 and returns UNKNOWN for absent or
   inconsistent state attributes, including S7-1200 responses that omit them.
--- a/s7commplus/connection.py
+++ b/s7commplus/connection.py
@@ -259,20 +259,37 @@ def _set_s7_groups(ctx: ssl.SSLContext) -> None:
     )
 
 
-def _verify_v3_hmac(protected: bytes, session_key: bytes) -> bytes:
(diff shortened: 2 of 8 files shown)


gijzelaerr:
This S7CommPlus work is now tracked in https://github.com/gijzelaerr/s7commplus/issues/15. Pull requests cannot be transferred between repositories, so this PR is being closed in favor of the new-repository tracking issue.

PR #887, opened 2026-09-11
fix(s7commplus): probe family-only session keys

### LINKED ISSUE #3: Support family-only fingerprints with legacy public-key fallback
## Problem

Some legacy PLCs return only a key-family identifier such as `00`, `01` or `03` instead of the complete `family:key-id` fingerprint.

Our parser requires the complete fingerprint, so authentication is skipped even when the correct family keys are present. The active .NET implementation now tries known family candidates on fresh connections and caches the successful key.

## Scope

- Distinguish complete fingerprints from family-only identifiers.
- Enumerate only public keys belonging to the reported family.
- Try each candidate on a fresh connection/session because a failed setup may invalidate the current one.
- Cache a confirmed mapping for subsequent connections.
- Provide an option to disable fallback where deterministic/no-probing behavior is required.
- Bound attempts and log key identifiers without exposing key material.

## Acceptance criteria

- Existing complete-fingerprint selection remains unchanged.
- Family-only responses can authenticate using a matching catalog key.
- A failed candidate is never retried in the same invalidated session.
- Unknown families and exhausted candidates fail clearly.
- Tests cover complete, family-only, unknown-family, disabled-fallback and cached-success paths.

## References

- Umbrella hardware issue: gijzelaerr/python-snap7#710
- DotNetProjects/S7CommPlusDriver automatic legacy public-key fallback

### LINKED ISSUE #834: Support family-only fingerprints with legacy public-key fallback
## Problem

Some legacy PLCs return only a key-family identifier such as `00`, `01` or `03` instead of the complete `family:key-id` fingerprint.

Our parser requires the complete fingerprint, so authentication is skipped even when the correct family keys are present. The active .NET implementation now tries known family candidates on fresh connections and caches the successful key.

## Scope

- Distinguish complete fingerprints from family-only identifiers.
- Enumerate only public keys belonging to the reported family.
- Try each candidate on a fresh connection/session because a failed setup may invalidate the current one.
- Cache a confirmed mapping for subsequent connections.
- Provide an option to disable fallback where deterministic/no-probing behavior is required.
- Bound attempts and log key identifiers without exposing key material.

## Acceptance criteria

- Existing complete-fingerprint selection remains unchanged.
- Family-only responses can authenticate using a matching catalog key.
- A failed candidate is never retried in the same invalidated session.
- Unknown families and exhausted candidates fail clearly.
- Tests cover complete, family-only, unknown-family, disabled-fallback and cached-success paths.

## References

- Umbrella hardware issue: #710
- DotNetProjects/S7CommPlusDriver automatic legacy public-key fallback

gijzelaerr:
Closes gijzelaerr/s7commplus#3

Some legacy PLCs advertise only a SessionKey family such as 00, 01, or 03. This adds a bounded fallback that:

- keeps complete-fingerprint selection unchanged
- enumerates only bundled keys from the reported family
- tries every candidate on a fresh transport/session
- treats setup rejection or setup-time disconnect as a rejected candidate
- caches a confirmed key by PLC host and port
- supports allow_legacy_key_fallback=False for deterministic no-probing operation
- keeps missing authentication dependencies as an immediate hard failure
- logs key identifiers, never key material

Validation:
- uv run pre-commit run --all-files
- uv run pytest -q --tb=short (2092 passed, 82 skipped)
- uv build --quiet

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,6 +100,12 @@ S7-1200/1500 PLCs. The ``s7`` package (recommended) and its ``snap7`` alias
 continue to work unchanged for legacy S7-300/400 PLCs and S7-1200/1500 with
 PUT/GET enabled.
 
+Some older PLCs advertise only their SessionKey family instead of a complete
+public-key fingerprint. The synchronous ``Client`` tries the bounded set of
+bundled keys from that family on fresh sessions and caches the confirmed key
+for the PLC. Set ``allow_legacy_key_fallback=False`` on ``connect()`` when key
+probing must be disabled.
+
 **Other new features in 4.0:**
 
 * **Command-line interface** (``s7 read``, ``s7 write``, ``s7 info``)
--- a/s7commplus/client.py
+++ b/s7commplus/client.py
@@ -29,7 +29,7 @@ from .codec import (
     encode_pvalue_typed,
     parse_create_object_session_id,
 )
-from .connection import S7CommPlusConnection
+from .connection import FamilyOnlyFingerprintError, S7CommPlusConnection, SessionKeyCandidateRejectedError
 from .protocol import DataType, ElementID, FunctionCode, Ids, ObjectId, ProtocolVersion
 from .subscription import (
     SubscriptionItem,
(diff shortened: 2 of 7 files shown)


gijzelaerr:
This S7CommPlus work is now tracked in https://github.com/gijzelaerr/s7commplus/issues/18. Pull requests cannot be transferred between repositories, so this PR is being closed in favor of the new-repository tracking issue.

PR #888, opened 2026-09-11
fix(s7commplus): renew legacy session keys

### LINKED ISSUE #835: Renew legacy S7CommPlus SessionKeys before PLC expiry
## Problem

Legacy integrity keys expire even while ordinary reads are active. The maintained .NET implementation renews them every 25 minutes by default; python-snap7 has no renewal mechanism, so a long-lived V1 connection will eventually fail.

Renewal must be performed under the existing authenticated session and the new key must not become active until the old-key-protected response is verified.

## Scope

- Add configurable legacy SessionKey refresh, enabled by default with a conservative interval.
- Perform the documented ServerSessionRequest / SecurityKey renewal exchange.
- Serialize renewal with application requests.
- Verify the renewal response with the old key before atomically switching to the new key.
- Stop/cancel refresh activity during disconnect and failure cleanup.
- Surface renewal failure predictably rather than silently continuing with an expired key.

## Acceptance criteria

- A deterministic short-interval test performs at least one renewal while reads continue.
- No request is encoded partly with the old key and partly with the new key.
- Failed renewal never installs the new key.
- Disconnect leaves no timer/thread/task behind.
- Sync behavior is implemented first; async support is tracked by the async parity issue.

## Dependencies

- Correct key descriptor encoding
- Atomic setup/key activation
- Response HMAC verification and request serialization

gijzelaerr:
Closes gijzelaerr/s7commplus#4

## Summary
- renew legacy SessionKeys every 25 minutes by default, with a configurable/disableable interval
- serialize the challenge and SecurityKey exchange with application requests
- keep the old key active through response-HMAC verification, then switch atomically
- close the connection and surface a stable error when renewal fails
- cancel renewal activity during disconnect and document the lifecycle

## Testing
- `pre-commit run --all-files`
- `uv run pytest -q --tb=short` (2104 passed, 82 skipped)
- `uv build --quiet`

## Dependency
This is stacked on gijzelaerr/python-snap7#884 because renewal must verify the PLC response with the old SessionKey before installing the new key.

### DIFF
--- a/s7commplus/client.py
+++ b/s7commplus/client.py
@@ -126,6 +126,7 @@ class S7CommPlusClient:
         tls_key: Optional[str] = None,
         tls_ca: Optional[str] = None,
         password: Optional[str] = None,
+        legacy_session_key_refresh_interval: Optional[float] = 25 * 60.0,
     ) -> None:
         """Connect to an S7-1200/1500 PLC using S7CommPlus.
 
@@ -139,6 +140,8 @@ class S7CommPlusClient:
             tls_key: Path to client private key (PEM)
             tls_ca: Path to CA certificate for PLC verification (PEM)
             password: PLC password for legitimation (V2+ with TLS)
+            legacy_session_key_refresh_interval: Seconds between legacy
+                SessionKey renewals, or ``None`` to disable them.
         """
         self._connect_params = {
             "host": host,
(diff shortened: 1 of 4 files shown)


gijzelaerr:
This S7CommPlus work is now tracked in https://github.com/gijzelaerr/s7commplus/issues/19. Pull requests cannot be transferred between repositories, so this PR is being closed in favor of the new-repository tracking issue.

PR #891, opened 2026-09-11
feat(s7commplus): complete subscription lifecycle

gijzelaerr:
Closes gijzelaerr/s7commplus#7.

Completes the sync and async subscription lifecycle with typed catalog-backed notifications, bounded per-subscription queues, callbacks/iterators, finite-credit replenishment, mixed alarm/data routing, deletion cleanup, reconnect cleanup, and observable loss diagnostics. Raw payload bytes remain available alongside decoded values.

This is stacked on gijzelaerr/python-snap7#886 for the symbolic tag catalog and includes the response-dispatch/integrity dependencies from gijzelaerr/python-snap7#883 and gijzelaerr/python-snap7#884.

Local verification:
- `uv run pre-commit run --all-files`
- `uv run --offline pytest -q --tb=short` (2123 passed, 82 skipped)
- `uv build --offline --quiet`

PR #886, opened 2026-09-11
feat(s7commplus): add typed symbolic tag API

gijzelaerr:
Closes gijzelaerr/s7commplus#6

Adds a public typed symbolic catalog and name-based sync/async APIs:

- preserve access path, software/wire datatype, SymbolCRC, array dimensions, string length, and optimized/classic offsets from browse metadata
- add resolve_tag, read_tag(s), and write_tag(s) with cached catalog lifecycle
- encode writes with the resolved PValue datatype instead of BLOB
- return per-item success/error results for batch operations
- refresh and safely retry failed reads once only when a fresh browse confirms the SymbolCRC changed
- never retry writes after a potentially ambiguous transport failure
- retain all existing low-level symbolic methods

Validation:
- uv run pre-commit run --all-files
- uv run pytest -q --tb=short (2085 passed, 82 skipped)
- uv build --quiet

PR #892, opened 2026-09-11
test: formalize classic S7 real PLC acceptance testing

### LINKED ISSUE #840: Formalize real-PLC acceptance testing with Gherkin scenarios
## Summary

Formalize volunteer testing against real Siemens PLCs as a repeatable acceptance-test program based on Cucumber/Gherkin scenarios, with results recorded as structured GitHub issues.

We already have real-hardware pytest coverage in `tests/test_client_e2e.py` (legacy S7) and `tests/test_s7_e2e.py` (S7CommPlus). The missing part is a stable, documented procedure and a queryable record of exactly which source revision was tested, on which PLC/firmware/configuration, by whom, when, and with what result.

This complements the protocol conformance suite from #620: conformance tests validate packets without hardware; this suite validates user-visible behaviour against real PLCs. It should also replace one-off reporting such as #793 with a reusable format.

## Goals

- Give volunteers one safe command and a small set of understandable scenarios.
- Reuse the existing pytest e2e implementation instead of maintaining a parallel test suite.
- Cover both legacy S7 and S7CommPlus, while allowing protocol-specific scenarios.
- Preserve enough environment and PLC metadata to reproduce failures.
- Make tested combinations and gaps discoverable through GitHub search.
- Produce useful diagnostics when a scenario fails without publishing PLC credentials, certificates, IP addresses, or proprietary programs.

### 1. Gherkin is the specification; pytest remains the runner

Add version-controlled `.feature` files under `tests/features/real_plc/`. Use `pytest-bdd` as the preferred Cucumber-compatible adapter because the project already uses pytest fixtures, markers, command-line options, JUnit output, and test helpers. Validate that choice with a small spike before converting all e2e coverage.

Step definitions should call shared helpers extracted from the current e2e tests. They must not duplicate protocol logic or invoke one test from another. Keep lower-level diagnostic tests in pytest where Gherkin would add no clarity.

Suggested tags:

- `@smoke`: read-only, safe to run on any correctly prepared test PLC
- `@write`: modifies only the dedicated scratch DB and always restores original bytes
- `@legacy_s7`: `snap7.Client`
- `@s7commplus`: sync S7CommPlus client
- `@async`: async client parity
- `@tls`, `@symbolic`, `@routing`, `@logo`: capability-specific scenarios
- `@administrative`: potentially disruptive operations; never part of the default run

### 2. Initial scenarios

```gherkin
Feature: Connect to and identify a real PLC

  Background:
    Given I am using a dedicated test PLC
    And the test configuration contains no secrets in reportable fields

  @smoke
  Scenario Outline: Establish and close a session
    Given the client uses the "<protocol>" protocol path
    When I connect to the configured PLC
    Then the client reports that it is connected
    And the negotiated protocol and security mode are recorded
    And the PLC order code, CPU family, firmware and CPU state are recorded when available
    When I disconnect
    Then the client reports that it is disconnected

    Examples:
      | protocol   |
      | legacy_s7 |
      | s7commplus |
```

```gherkin
Feature: Read a known test data block

  @smoke
  Scenario: Read the canonical fixture
    Given the read-only test DB has the documented canonical layout
    When I read the complete fixture DB
    Then INT, REAL, BYTE, WORD, DWORD, DINT, CHAR and BOOL values match the fixture
    And individual reads return the same values as the complete-block read

  @smoke
  Scenario: Read multiple values in one request
    Given the read-only test DB has the documented canonical layout
    When I read values of different sizes in one multi-variable request
    Then every value and result code matches the fixture
```

```gherkin
Feature: Write safely to a scratch data block

  @write
  Scenario Outline: Round-trip a value and restore the PLC
    Given writing has been explicitly enabled
    And I have a dedicated scratch DB
    And the original bytes at the target address have been saved
    When I write a valid <type> value
    Then reading the address returns the written value
    And the original bytes are restored even if the scenario fails

    Examples:
      | type  |
      | INT   |
      | REAL  |
      | BYTE  |
      | WORD  |
      | DWORD |
      | DINT  |
      | CHAR  |
      | BOOL  |
```

```gherkin
Feature: Recover from normal connection lifecycle events

  @smoke
  Scenario: Reconnect after a clean disconnect
    Given I connected to and disconnected from the PLC
    When I reconnect with the same configuration
    Then a known read succeeds

  @smoke
  Scenario: Repeated operations do not corrupt the session
    Given I am connected to the PLC
    When I read the canonical fixture repeatedly
    Then every read succeeds with the expected value
    And disconnect completes cleanly
```

Protocol-specific follow-ups should cover S7CommPlus authentication/TLS negotiation, symbolic browse/read/write, optimized versus non-optimized DB behaviour, legacy I/Q/M access, routing, LOGO, and sync/async parity. Those scenarios should be skipped with a machine-readable reason when the PLC does not advertise or is not configured for the capability; unsupported is not the same as failed.

### 3. Test tiers and safety

The default command runs only `@smoke` scenarios and is read-only. Write scenarios require both an explicit flag (for example `--allow-plc-write`) and a configured scratch DB. Before each write, save the affected bytes; restore them in fixture teardown/finally logic and verify restoration.

PLC start/stop, clock changes, block upload/download, protection changes, and other operational actions belong in `@administrative`. They require a separate explicit opt-in and a dedicated non-production PLC. They must never run as part of the normal volunteer procedure.

Provide a canonical TIA Portal fixture project or precise setup document for the existing 37-byte DB layout. Document optimized-block requirements, PUT/GET requirements, rack/slot defaults, and least-privilege/security prerequisites. Never ask testers to weaken a PLC that is in service.

### 4. Reproducible report

Each run should emit:

- a human-readable terminal summary;
- JUnit XML for scenario-level pass/fail/skip results;
- a small JSON report with a versioned schema.

The JSON report should include:

- test-suite schema version, UTC start/end time, and overall result;
- exact git commit SHA and whether the worktree was dirty (or exact installed package version and source);
- Python version, OS, architecture, and relevant optional dependencies;
- tester's GitHub handle (supplied explicitly or taken from the report issue author);
- PLC family/model/order code, firmware, rack/slot, protocol path/version, TLS/security mode, and relevant TIA configuration;
- scenario IDs/tags and pass/fail/skip status;
- sanitized exception, traceback and logs for failures.

Do not record PLC IPs, passwords, private keys, certificate private material, plant names, or other network/site identifiers. Packet captures should be optional, manually reviewed, sanitized, and attached only when needed.

### 5. GitHub-native tracking

Add an issue form such as `.github/ISSUE_TEMPLATE/plc-test-result.yml` for one issue per tester + PLC configuration + source revision + run. The form should collect the report fields above and require the generated summary/JSON attachment. The GitHub issue author and creation timestamp provide an auditable "who" and "when"; the explicit UTC run time covers delayed submissions.

Use a small, stable label set rather than labels for every firmware/Python/OS combination:

- `hardware-test`
- `test-result: pass`, `test-result: fail`, or `test-result: partial`
- `protocol: legacy-s7` or `protocol: s7commplus`

Keep high-cardinality data (exact PLC model, firmware, Python version, OS, commit SHA) in the structured report so it can be parsed and searched without label sprawl.

Add a saved issue-search link to the documentation for each useful view, for example open failures, S7CommPlus runs, and results for a release/commit. Use this issue as the implementation tracker; create/link individual result issues rather than accumulating free-form comments here. GitHub Discussions can coordinate volunteers, but test evidence belongs in issues so it remains structured and queryable. Do not require GitHub Projects, since Projects are currently disabled for this repository.

Optionally add a GitHub Action triggered for `hardware-test` issues to validate the JSON schema and post a normalized summary. It must not treat self-reported hardware results as a required PR check. A later action can generate a compatibility matrix from valid result issues; the issues remain the source of truth.

## Coverage strategy

Track coverage across these dimensions:

1. Client/protocol: legacy S7, S7CommPlus sync, S7CommPlus async, LOGO, routing where available.
2. PLC family/model: S7-200, S7-300, S7-400, S7-1200, S7-1500, LOGO, with exact order code.
3. Firmware and security mode: firmware version, PUT/GET, optimized blocks, access protection, TLS/authentication mode.
4. Host: Python 3.10-3.14, Windows/Linux/macOS, architecture.
5. Source: release version plus immutable commit SHA.

Prioritize PLC model/firmware/protocol coverage over exhaustively multiplying host OS and Python versions; CI already covers the host matrix against emulated/local tests. For a release candidate, define a small required hardware baseline and record gaps explicitly. A result becomes stale when the tested code, scenario schema, relevant protocol implementation, or PLC firmware/configuration changes—not merely because time passed.

## Deliverables / acceptance criteria

- [ ] Decide/prototype the Cucumber adapter; prefer `pytest-bdd` unless the spike identifies a concrete blocker.
- [ ] Add versioned Gherkin feature files for connection, fixture reads, multi-read, safe write/read/restore, and reconnect lifecycle.
- [ ] Refactor the existing e2e tests so Gherkin steps and diagnostic pytest tests share fixtures/helpers without duplicated protocol logic.
- [ ] Add a read-only default command and explicit write/admin opt-ins.
- [ ] Guarantee and test scratch-data restoration on success, assertion failure, exception, and interruption where process cleanup is possible.
- [ ] Add canonical PLC fixture setup and volunteer runbook documentation.
- [ ] Generate JUnit XML plus schema-versioned JSON containing source, environment, PLC metadata, and scenario results.
- [ ] Add the `plc-test-result` GitHub issue form and the small stable label set.
- [ ] Document saved searches and the policy for pass/fail/partial, skip reasons, reruns, superseded results, and stale results.
- [ ] Validate the procedure with at least one legacy S7 PLC and one S7CommPlus-capable S7-1200/1500, filing result issues for both.
- [ ] Keep real-PLC tests opt-in and out of normal hosted CI; do not add an internet-exposed or untrusted self-hosted PLC runner as part of this issue.

## Open questions

- Should the canonical fixture be distributed as a small TIA Portal project, a source export, or both?
- Which exact PLC/firmware combinations form the minimum release-candidate baseline?
- Should a generated compatibility matrix live in the documentation, a release checklist issue, or both?

gijzelaerr:
Progresses #840.

Adds a pytest-bdd-based real-PLC acceptance program with 13 versioned scenarios covering connection, canonical reads, multi-read, reconnect behavior, and safely restored scratch writes using the classic S7 protocol.

Safety and evidence:
- smoke-only default runner; writes/admin require separate explicit opt-ins
- shared classic-S7 helpers used by the existing diagnostic e2e suite
- scratch bytes restored and read-back verified from fixture teardown
- Gherkin terminal output, JUnit XML, and schema-versioned sanitized JSON
- issue form, saved searches, fixture setup, and volunteer runbook
- real-PLC tests remain opt-in and outside hosted CI

Local verification:
- `uv run --no-sync pre-commit run --all-files`
- `uv run --no-sync pytest` (1385 passed, 63 skipped)
- `uv build --offline`
- `uv run --no-sync sphinx-build -N -W -b html doc ...`

The final acceptance item—filing a real-hardware result—requires a volunteer PLC run after this infrastructure lands, so this PR intentionally does not close #840.

### DIFF
--- a/.github/ISSUE_TEMPLATE/plc-test-result.yml
+++ b/.github/ISSUE_TEMPLATE/plc-test-result.yml
@@ -1,0 +1,78 @@
+name: Real PLC test result
+description: Record one reproducible python-snap7 acceptance run against a real PLC
+title: "[hardware test]: "
+labels: ["hardware-test"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Attach the generated JSON and JUnit XML. Review all attachments before uploading. Never include PLC addresses, credentials, private keys, plant names, or unreviewed packet captures.
+  - type: input
+    id: source
+    attributes:
+      label: Source revision
+      description: Immutable 40-character commit SHA and, if applicable, release version.
+      placeholder: abcdef0123456789abcdef0123456789abcdef01
+    validations:
+      required: true
+  - type: dropdown
+    id: result
+    attributes:
+      label: Overall result
+      options: [pass, fail, partial]
+    validations:
+      required: true
+  - type: dropdown
+    id: protocol
+    attributes:
+      label: Protocol path
+      options: [legacy-s7, s7commplus]
+    validations:
+      required: true
+  - type: input
+    id: plc
+    attributes:
+      label: PLC family, model, and order code
+      placeholder: S7-1500, CPU 1511-1 PN, 6ES7 ...
+    validations:
+      required: true
+  - type: input
+    id: firmware
+    attributes:
+      label: Firmware and security mode
+      placeholder: V2.9, PUT/GET enabled, non-optimized fixture DBs
+    validations:
+      required: true
+  - type: input
+    id: host
+    attributes:
+      label: Host environment
+      placeholder: Python 3.13, Linux x86_64
+    validations:
+      required: true
+  - type: input
+    id: started
+    attributes:
+      label: Run time (UTC)
+      placeholder: 2026-09-11T12:34:56Z
+    validations:
+      required: true
+  - type: textarea
+    id: artifacts
+    attributes:
+      label: Sanitized report artifacts
+      description: Drag the generated JSON and JUnit XML here. State any manually reviewed capture separately.
+    validations:
+      required: true
+  - type: textarea
+    id: skips
+    attributes:
+      label: Skips, failures, and notes
+      description: Include machine-readable skip reasons and whether this result supersedes an earlier issue.
+  - type: checkboxes
+    id: safety
+    attributes:
+      label: Safety confirmation
+      options:
+        - label: I reviewed the attachments and they contain no PLC address, credentials, private keys, plant/site identifiers, or proprietary PLC program.
+          required: true
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ CHANGES
 
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
+* Add a safe, Gherkin-based real-PLC acceptance program with restored scratch
+  writes and sanitized JSON/JUnit reports (#840).
+
 * Decode corroborating CPU execution attributes so S7CommPlus `get_cpu_state()`
   distinguishes RUN from STOP on S7-1500 and returns UNKNOWN for absent or
   inconsistent state attributes, including S7-1200 responses that omit them.
--- a/README.rst
+++ b/README.rst
@@ -27,6 +27,8 @@ As of version 3.0, the C library is no longer used, but the name is kept for
 backwards compatibility.
 
 The full documentation is available on `Read The Docs <https://python-snap7.readthedocs.io/en/latest/>`_.
+The safe volunteer hardware-test procedure is in
+`Real-PLC acceptance testing <doc/real-plc-testing.rst>`_.
 
 
 Quick Start
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -34,6 +34,7 @@ Welcome to python-snap7's documentation!
    :caption: Development
 
    development
+   real-plc-testing
 
 .. toctree::
    :maxdepth: 2
--- a/doc/real-plc-testing.rst
+++ b/doc/real-plc-testing.rst
@@ -1,0 +1,90 @@
+Real-PLC acceptance testing
+===========================
+
+The versioned Gherkin specifications in ``tests/features/real_plc`` exercise the
+same client APIs and canonical 37-byte fixture used by the diagnostic end-to-end
+tests. ``pytest-bdd`` was selected because it keeps pytest as the runner, reuses
+fixtures, maps tags to markers, and supports the existing JUnit workflow.
+
+Safety model
+------------
+
+Use only a dedicated, non-production test PLC. The default runner selects
+``@smoke`` and is read-only. It never enables writes or administrative actions.
+Scratch writes require ``--allow-write``, a separate DB (DB2 by default), and are
+wrapped in a guard that saves, restores, and reads back the original bytes from
+fixture teardown even when a step fails or pytest handles an interruption.
+Hard process termination or power loss cannot run cleanup, so inspect the scratch
+DB before reuse after either event.
+
+Administrative actions (CPU start/stop, clock changes, program transfers, or
+protection changes) require ``--allow-admin`` and must never target an in-service
+PLC. No administrative scenarios are included in the initial suite.
+
+PLC preparation
+---------------
+
+Import ``tests/plc_setup/e2e_test_dbs.scl`` into TIA Portal, or recreate its exact
+layout. It defines read-only DB1 and scratch DB2. On S7-1200/1500, disable
+``Optimized block access`` for both fixture DBs. Enable PUT/GET only when the
+chosen legacy protocol path and the PLC security policy require it. Defaults are
+rack 0, slot 1, and TCP port 102. Apply the least privilege that permits the
+selected scenarios; do not weaken a PLC that is in service.
+
+Running and reporting
+---------------------
+
+Install the test extras, then run the safe profile. Values supplied to reportable
+metadata must not contain addresses, credentials, certificate paths, plant names,
+or other site identifiers::
+
+  uv sync --extra test --extra s7commplus
+  uv run python tools/run_real_plc_acceptance.py \
+    --plc-ip YOUR_PRIVATE_ADDRESS \
+    --protocol legacy_s7 \
+    --tester @YOUR_HANDLE \
+    --plc-family S7-1500 \
+    --plc-model "CPU 1511-1 PN" \
+    --plc-firmware V2.9 \
+    --plc-security-mode "PUT/GET, non-optimized fixture DBs"
+
+The runner prints a Gherkin-aware terminal summary and writes JUnit XML plus a
+schema-versioned, sanitized JSON report beneath ``real-plc-results/``. The JSON
+records exact source state, host environment, PLC metadata, scenario tags,
+pass/fail/skip results, and bounded diagnostics. It deliberately never reads the
+PLC address into the report. Review both artifacts before publishing them.
+
+Add ``--allow-write`` only after confirming DB2 is disposable scratch space.
+``--allow-admin`` remains separate so write permission never implies permission
+for operational actions.
+
+Result policy
+-------------
+
+File one ``Real PLC test result`` issue per tester, PLC configuration, source
+revision, and run. Attach both generated artifacts and apply exactly one result
+label and one protocol label:
+
+* ``hardware-test``
+* ``test-result: pass``, ``test-result: fail``, or ``test-result: partial``
+* ``protocol: legacy-s7`` or ``protocol: s7commplus``
+
+``pass`` means every selected applicable scenario passed. ``fail`` means at least
+one selected scenario failed. ``partial`` means the run completed with capability
+skips or an incomplete selection; unsupported capabilities are not failures.
+Reruns get a new issue and link the earlier result. Close the earlier issue as
+superseded only after the replacement artifacts exist. A result becomes stale
+when the tested code, feature schema, relevant protocol implementation, PLC
+firmware, or PLC configuration changes—not merely with age.
+
+Useful searches:
+
+* `Open hardware failures <https://github.com/gijzelaerr/python-snap7/issues?q=is%3Aissue+is%3Aopen+label%3Ahardware-test+label%3A%22test-result%3A+fail%22>`_
+* `S7CommPlus hardware runs <https://github.com/gijzelaerr/python-snap7/issues?q=is%3Aissue+label%3Ahardware-test+label%3A%22protocol%3A+s7commplus%22>`_
+* Search a release or commit by adding its version or SHA to ``is:issue label:hardware-test``.
+
+For a release candidate, prioritize one supported legacy S7 PLC and one
+S7CommPlus-capable S7-1200/1500 before expanding the host matrix; hosted CI already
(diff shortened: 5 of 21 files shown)


gijzelaerr:
@russwing Could you run the cross-model real-PLC acceptance pass for the S7CommPlus work once its code review settles? Your S7-1200 and S7-1500 access makes that the useful compatibility pair. Suggested order: the response/authentication roots (#881, #882, #883, #887), then stacked integrity/renewal (#884, #888), followed by typed tags/subscriptions (#886, #891). Please use the result template in this PR and record the exact CPU, firmware, protection mode, and tested commit. #881 still has an unresolved S7-1511C item-read failure from the latest reporter retest, so a failure there would also be valuable evidence.

russwing:
Ran the harness today on both CPUs of the compatibility pair. **One blocking defect in the S7CommPlus adapter, and a preparation gap of ours.** Commit `eb6d9e3`.

## The blocker: the S7CommPlus adapter never enables TLS

`tests/real_plc/support.py::S7CommPlusAdapter.connect` calls:

```python
self.client.connect(self.config.host, self.config.port, self.config.rack, self.config.slot)
```

`S7CommPlusClient.connect` defaults `use_tls: bool = False`, and its own docstring says *"required for V2"*. **S7CommPlus V2 is mandatory-TLS** — S7-1200 firmware V4.5+ and S7-1500 V2.x+ — so the adapter cannot reach any modern CPU. Every `s7commplus` scenario fails at connect with `S7ConnectionError: Connection closed by peer`, which is the RST an unencrypted V1 SetupSession earns from a V2 target.

**Controlled proof** — same CPU, same commit, same port, only `use_tls` differs:

```
S7-1200 1212C V4.7.3     use_tls=False -> S7ConnectionError: Connection closed by peer
S7-1200 1212C V4.7.3     use_tls=True  -> CONNECTED protocol_version=V2
S7-1500 1511F V2.9.7     use_tls=False -> S7ConnectionError: Connection closed by peer
S7-1500 1511F V2.9.7     use_tls=True  -> CONNECTED protocol_version=V2
```

**Suggested fix:** a `--plc-use-tls` flag (plus `--plc-tls-cert/key/ca` passthrough) on `run_real_plc_acceptance.py`, threaded into the adapter, and recorded in the report metadata since it is part of the tested security mode. Without it the acceptance program can only certify pre-V4.5 S7-1200s and V1 S7-1500s.

## Results as they stand

| protocol | S7-1200 1212C V4.7.3 | S7-1500 1511F-1 PN V2.9.7 |
|---|---|---|
| `legacy_s7` | **Establish and close a session — PASS** | **Establish and close a session — PASS** |
| `s7commplus` | all fail at connect (no TLS) | all fail at connect (no TLS) |

The other four scenarios fail on both CPUs for a reason that is **ours, not yours**: neither PLC carries the DB1/DB2 fixture from `tests/plc_setup/e2e_test_dbs.scl`. Importing it needs TIA Portal and our automation engineer; it is requested and should land with his next session, after which I can run the full suite including `--allow-write`.

So this is a **partial**, and I would rather report it now than sit on the TLS finding.

## One note on hardware scope

#881's outstanding item-read failure is on an **S7-1511C**. Ours is a **1511F** — same S7-1500 family, different model and a safety variant. Whatever we produce is useful compatibility evidence but will neither confirm nor clear that specific report.

I will file the formal `Real PLC test result` issues with both artifacts once the fixture DBs are in and the TLS path is reachable — filing a result now would mostly record a harness gap rather than PLC behaviour.

russwing:
Fixture DBs are now in place on both CPUs, so here is the full pass. **Two defects, one of them substantial**, plus confirmation that the TLS fix works.

Harness `eb6d9e3`, plus a one-line local patch described below. Both CPUs are dedicated bench units.

## Results

| | S7-1200 1212C V4.7.3 | S7-1500 1511F-1 PN V2.9.7 |
|---|---|---|
| **legacy_s7** smoke | **5/5 PASS** | 1/5 — PUT/GET is disabled (`class=0x81 code=0x04`, service not implemented). A CPU setting, not a library issue. |
| **s7commplus** smoke, stock | 0/5 — cannot connect | 0/5 — cannot connect |
| **s7commplus** smoke, TLS patched | 1/5 — session PASSES, all reads fail | 1/5 — session PASSES, all reads fail |

## 1. The TLS fix works — confirmed

Patching `S7CommPlusAdapter.connect` to pass `use_tls=True` turns the connection scenario from fail to **PASS on both CPUs**. That confirms the fix proposed in my earlier comment. Please add `--plc-use-tls` (plus cert/key/ca passthrough) and record it in the report metadata.

## 2. `db_read` over S7CommPlus ignores offset and size

This is the bigger one, and it is **not** fixed by #881 — I merged #881 into the harness branch and re-ran; byte-for-byte identical failure on both CPUs.

Same CPU (1212C), same DB, same offsets, both protocol paths side by side:

```
DB1.0  size= 2   legacy=00 0a                     s7cp=92 00 00 01
DB1.0  size= 8   legacy=00 0a 00 ff 42 f6 e6 66   s7cp=92 00 00 01
DB1.2  size= 4   legacy=00 ff 42 f6               s7cp=42
DB1.0  size=16   legacy=<16 correct bytes>        s7cp=92 00 00 01
```

Three separate problems visible in four lines:

- **Size is ignored.** Requests for 2, 8 and 16 bytes all return the same four bytes.
- **The data does not match the block.** `92 00 00 01` appears nowhere in DB1's real contents.
- **Offset is ignored or mis-scaled.** `DB1.2 size=4` returns a single byte `42`, which is the byte at DB1.**4**.

The legacy column is ground truth from the same CPU seconds apart, so the fixture and the DB are demonstrably correct.

**Identical bytes on two different CPU families** (1212C V4.7.3 and 1511F V2.9.7) points at client-side decode rather than anything CPU-specific.

## On #881

You flagged an unresolved S7-1511C item-read failure and said a failure would be useful evidence. We reproduce a read failure on a **1511F** and a **1212C**, and **#881 as it stands does not resolve either**. Whether it is the same root cause as the 1511C report I cannot say — different model, and I do not have that reporter's fixture.

Given the size/offset behaviour above, I would treat this as distinct from "unprefixed response values": the transport is not returning a wrong-shaped version of the right data, it is returning the wrong data.

## Scope notes

- **No `--allow-write` run yet.** DB2 exists and writes are cleared on these bench units, but running scratch writes over a path whose reads are this wrong would tell us little and risks writing somewhere unintended. Happy to run writes once reads are trustworthy.
- I have **not** filed formal `Real PLC test result` issues yet — the current results mostly characterise two library defects rather than PLC behaviour, and I would rather file a clean result once the TLS flag lands and reads are fixed. Say if you would prefer them filed as-is and I will.
- PUT/GET on the 1500 can be enabled if you want the legacy path covered on that family; it is off by default there and the runbook rightly says not to weaken a PLC without cause.

gijzelaerr:
The S7CommPlus portion has now moved to the standalone repository in https://github.com/gijzelaerr/s7commplus/pull/30; this PR remains the Classic S7 acceptance work.

That port addresses both findings from the hardware pass:

- TLS is configurable with `--plc-use-tls` and certificate/key/CA passthrough, including the existing E2E path. Reports capture negotiated TLS plus boolean client-certificate/CA usage without exposing paths or the PLC address.
- Raw DB access now sends the ClassicBlob marker, a zero-based byte offset, and an independent byte size. BLOB PValues include BlobRootId, and raw writes use the BLOB wire type. The exact paths and values are covered by byte-level and emulator regressions for sync/async clients.
- The canonical DB1/DB2 SCL fixture is included in that repository, so fixture preparation is no longer external to the test source.

The standalone PR merged with all hosted checks green (805 passed, 43 hardware-gated skips locally, plus typing, lint/format, docs, package/Twine, lock, and pre-commit checks). I have deliberately not treated unit/emulator coverage as proof on hardware. The next useful step is to rerun S7CommPlus master on the same 1212C V4.7.3 and 1511F V2.9.7, first smoke and then `--allow-write` only after reads return the canonical bytes.

PR #893, opened 2026-09-11
fix(server): harden COTP connection framing

### LINKED ISSUE #889: Pure-Python server: Connection Confirm is still rejected by Snap7-derived clients — TSAP parameters are not echoed
## Environment

- python-snap7 3.1.2 (PyPI, released 2026-08-11) and current `master`
- Pure-Python server (`snap7.server.Server`), Windows 11, CPython 3.14
- Client: Telegraf 1.30.3 `inputs.s7comm`, which uses [gos7](https://github.com/robinson/gos7)

## Summary

#813 / #816 fixed the extra reserved byte in the Connection Confirm, and the
framing is correct on `master` now. The CC is still **14 bytes** though, because
the server does not echo the TSAP parameters from the Connection Request.
Snap7 and its ports accept **only a 22-byte** Connection Confirm, so any
Snap7-derived client still fails to connect — with `master` just as with 3.1.2.

The library's own client is unaffected: it does not check the length, which is
probably why this has gone unnoticed. The same goes for Apache PLC4X, which #816
was verified against - it accepts the 14-byte frame, so that test could not have
caught it. Snap7 and its ports are stricter.

## Why 22 bytes

`gos7/tcpclient.go`, `isoConnect()`:

```go
response, err := mb.Send(msg)
if size := len(response); size == 22 {
    if mb.LastPDUType != byte(0xD0) { // 0xD0 = CC Connection confirm
        err = fmt.Errorf("errIsoConnect")
    }
} else {
    err = fmt.Errorf(ErrorText(errIsoInvalidPDU))
}
```

The length is compared for equality, not for a minimum. This is inherited from
Snap7's C implementation, so the same applies to Snap7 itself and to the other
ports. A real S7-300/400 answers the Connection Request by echoing both TSAP
parameters, which is exactly what makes the frame 22 bytes.

## Reproducer

Send a standard Connection Request and dump the answer — no PLC and no client
library needed:

```python
import socket, struct

cr = bytes([0x11, 0xE0, 0x00, 0x00, 0x00, 0x01, 0x00,
            0xC0, 0x01, 0x0A,        # TPDU size 2**10
            0xC1, 0x02, 0x01, 0x00,  # source TSAP
            0xC2, 0x02, 0x01, 0x02]) # destination TSAP (rack 0 / slot 2)
s = socket.create_connection(("127.0.0.1", 1102), timeout=5)
s.sendall(struct.pack(">BBH", 3, 0, 4 + len(cr)) + cr)
head = s.recv(4)
body = s.recv(struct.unpack(">H", head[2:])[0] - 4)
print(len(head) + len(body), (head + body).hex(" "))
```

Results against a server with `register_area` for PE/PA/MK/DB:

| version | bytes | frame |
| --- | --- | --- |
| 3.1.2 | 15 | `03 00 00 0f 09 d0 00 00 01 00 01 00 c0 01 0a` |
| `master` (with #816) | 14 | `03 00 00 0e 09 d0 00 01 00 01 00 c0 01 0a` |
| real PLC / expected | 22 | `03 00 00 16 11 d0 00 01 00 01 00 c0 01 0a c1 02 01 00 c2 02 01 02` |

With 14 bytes, Telegraf reports:

```
connecting to "127.0.0.1:1102" failed: ISO : Invalid PDU received
```

for every rack/slot combination, while `snap7.client.Client().connect(...)`
succeeds against the same server at the same moment.

## Suggested fix

Keep the TSAPs from the Connection Request and mirror them back. Sketch,
against `master`:

```python
def _parse_cotp_cr(self, data: bytes) -> bool:
    ...
    self._cr_src_tsap = self._cr_dst_tsap = None
    offset = 7
    while offset + 2 <= len(data):
        code, length = data[offset], data[offset + 1]
        if offset + 2 + length > len(data):
            break
        if code == 0xC1:
            self._cr_src_tsap = bytes(data[offset + 2:offset + 2 + length])
        elif code == 0xC2:
            self._cr_dst_tsap = bytes(data[offset + 2:offset + 2 + length])
        offset += 2 + length
    ...

def _build_cotp_cc(self) -> bytes:
    params = struct.pack(">BBB", self.COTP_PARAM_PDU_SIZE, 1, self.tpdu_size)
    for code, tsap in ((0xC1, self._cr_src_tsap), (0xC2, self._cr_dst_tsap)):
        if tsap:
            params += struct.pack(">BB", code, len(tsap)) + tsap
    header = struct.pack(">BBHHB", 6 + len(params), self.COTP_CC,
                         self.dst_ref, self.src_ref, 0x00)
    return header + params
```

Mirroring what the client asked for also keeps the server honest about which
TSAP the connection was accepted on. With this applied, Telegraf reads PE, PA,
MK and DB areas from the pure-Python server (INT, REAL, BOOL and DINT verified),
and `snap7.client` keeps working unchanged.

I am happy to open a PR with this plus a regression test that asserts the CC is
22 bytes and contains both TSAP parameters.

The bug is found and the issue is created by Claude Opus 5

### LINKED ISSUE #890: Pure-Python server: RECEIVE_DEADLINE expiring mid-frame desynchronises the connection
## Environment

- python-snap7 3.1.2 and current `master` (4984721)
- Pure-Python server (`snap7.server.Server`), Windows 11, CPython 3.14
- Client: Telegraf 1.30.3 `inputs.s7comm` (gos7)

## Summary

The absolute receive deadline added in #810 is a sensible bound on how long a request may take once it has started. But it also runs while the connection is **idle**, and when it expires between the TPKT header and the payload of the same frame, the header is consumed and the rest of the frame is left in the stream. The next read then starts in the middle of that frame:

```
Error handling client ('172.18.0.3', 58258): Invalid TPKT version: 2
```

`0x02` is the first byte of the COTP DT header (`02 f0 80`), i.e. the read resumed four bytes into the frame. The client sees `connection reset by peer`.

## Mechanism

In `ServerISOConnection.receive_data()` the deadline is taken once, before the first byte, and used for both reads:

```python
deadline = time.monotonic() + self.RECEIVE_DEADLINE
while True:
    tpkt_header = self._recv_exact(4, deadline)        # waits while idle
    ...
    payload = self._recv_exact(remaining, deadline)    # same budget
```

A client that sends its next request shortly before the deadline gets its TPKT header read with almost no budget left. `_recv_exact` checks the deadline **before** it calls `recv()`, so the payload read raises a plain `TimeoutError` even though the payload is typically already sitting in the socket buffer - it arrives in the same TCP segment. Because that call itself read zero bytes, the "partial frame" guard in `_recv_exact` does not trigger.

The client handler then treats it as an idle timeout:

```python
except socket.timeout:      # alias of TimeoutError
    continue
```

The header is gone, the payload is not, and the next `receive_data()` parses `02 f0 80 32` as a TPKT header.

## What makes it visible

Only a polling interval close to the deadline triggers it. A client polling faster always sends well before the deadline; one polling slower hits a clean timeout before the first byte, which is handled correctly. Measured against Telegraf, 60 s per run, 8 fields, otherwise identical config:

| poll interval | aborted connections in 60 s |
| --- | --- |
| 2 s | 0 |
| **5 s** | **8** |
| 6 s | 0 |

So the failure window sits exactly at `RECEIVE_DEADLINE`, which makes it look intermittent and hard to attribute. A 5 s polling interval is not unusual for data collection.

## Suggested fix

Keep the bound from #810, but start it when a frame starts rather than while waiting for one:

```python
tpkt_header = self._recv_exact(4, idle_deadline)   # idle wait, may time out cleanly
deadline = time.monotonic() + self.RECEIVE_DEADLINE
...
payload = self._recv_exact(remaining, deadline)
```

The idle wait is already handled by the `socket.timeout` / `continue` loop, and a slow sender still gets cut off once it has begun a frame, so the protection against slow or stalled requests stays intact.

Independently of that, a timeout **after** the header has been consumed leaves the stream in an undefined state. Raising `S7ConnectionError` there, rather than a `TimeoutError` the handler swallows, would make any remaining case fail loudly instead of desynchronising.

Locally I worked around the symptom with `RECEIVE_DEADLINE = 300`, after which the 5 s poll runs clean (0 aborts, 0 errors in 60 s). That gives up the bound from #810 though, so I would not propose it as the fix.

Happy to turn this into a PR if the direction sounds right.

The bug is found and thus issue is create by Claude Code Opus 5.

gijzelaerr:
## Summary

- echo calling and called TSAP parameters in COTP Connection Confirm frames for Snap7-derived clients
- start a fresh receive deadline after each complete TPKT header
- treat payload timeouts as connection errors to prevent stream desynchronisation
- add exact wire-format and deadline regression coverage

## Verification

- `uv run pytest` — 2076 passed, 82 skipped
- `uv run pre-commit run --all-files`
- `uv build`

Closes #889
Closes #890

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ CHANGES
 
 Major release: new `s7commplus` package with S7CommPlus protocol support.
 
+* Echo calling and called TSAP parameters in pure-Python server connection
+  confirmations, and restart the receive deadline after each TPKT header.
 * Decode corroborating CPU execution attributes so S7CommPlus `get_cpu_state()`
   distinguishes RUN from STOP on S7-1500 and returns UNKNOWN for absent or
   inconsistent state attributes, including S7-1200 responses that omit them.
--- a/snap7/server/__init__.py
+++ b/snap7/server/__init__.py
@@ -2676,8 +2676,10 @@ class ServerISOConnection:
     COTP_DC = 0xC0  # Disconnect Confirm
     COTP_DT = 0xF0  # Data Transfer
 
-    # COTP parameter code for TPDU size (ISO 8073)
+    # COTP parameter codes (ISO 8073)
     COTP_PARAM_PDU_SIZE = 0xC0
+    COTP_PARAM_CALLING_TSAP = 0xC1
+    COTP_PARAM_CALLED_TSAP = 0xC2
 
     def __init__(self, client_socket: socket.socket):
         """Initialize server ISO connection."""
@@ -2687,6 +2689,8 @@ class ServerISOConnection:
         self.src_ref = 0x0001  # Server reference
         self.dst_ref = 0x0000  # Client reference (assigned during handshake)
         self.tpdu_size = 0x0A  # Default: 1024 bytes (2^10)
+        self.calling_tsap: bytes | None = None
+        self.called_tsap: bytes | None = None
 
     def accept_connection(self) -> bool:
         """Accept ISO connection from client."""
@@ -2735,9 +2739,9 @@ class ServerISOConnection:
         """
         fragments: list[bytes] = []
         total_size = 0
-        deadline = time.monotonic() + self.RECEIVE_DEADLINE
         while True:
-            tpkt_header = self._recv_exact(4, deadline)
+            header_deadline = time.monotonic() + self.RECEIVE_DEADLINE
+            tpkt_header = self._recv_exact(4, header_deadline)
             version, reserved, length = struct.unpack(">BBH", tpkt_header)
 
             if version != 3:
@@ -2747,7 +2751,11 @@ class ServerISOConnection:
             if remaining <= 0:
                 raise S7ConnectionError("Invalid TPKT length")
 
-            payload = self._recv_exact(remaining, deadline)
+            frame_deadline = time.monotonic() + self.RECEIVE_DEADLINE
+            try:
+                payload = self._recv_exact(remaining, frame_deadline)
+            except TimeoutError as e:
+                raise S7ConnectionError("Receive deadline exceeded after TPKT header") from e
 
             if len(payload) < 3:
                 raise S7ConnectionError("Invalid COTP DT: too short")
@@ -2802,18 +2810,25 @@ class ServerISOConnection:
         # Store client reference
         self.dst_ref = src_ref
 
-        # Parse variable parameters for TPDU size
+        # Parse variable parameters used in the connection confirmation.
+        self.calling_tsap = None
+        self.called_tsap = None
         offset = 7
         while offset + 2 <= len(data):
             param_code = data[offset]
             param_len = data[offset + 1]
             if offset + 2 + param_len > len(data):
                 break
+            param_data = data[offset + 2 : offset + 2 + param_len]
             if param_code == self.COTP_PARAM_PDU_SIZE and param_len == 1:
                 exponent = data[offset + 2]
                 if 7 <= exponent <= 13:
                     self.tpdu_size = exponent
                     logger.debug(f"Client requested TPDU size 2^{exponent} = {1 << exponent}")
+            elif param_code == self.COTP_PARAM_CALLING_TSAP:
+                self.calling_tsap = param_data
+            elif param_code == self.COTP_PARAM_CALLED_TSAP:
+                self.called_tsap = param_data
             offset += 2 + param_len
 
         logger.debug(f"Received COTP CR from client ref {src_ref}")
@@ -2826,8 +2841,15 @@ class ServerISOConnection:
         negotiated maximum segment size and don't fall back to the
         ISO 8073 class-0 default of 128 bytes.
         """
-        pdu_size_param = struct.pack(">BBB", self.COTP_PARAM_PDU_SIZE, 1, self.tpdu_size)
-        pdu_length = 6 + len(pdu_size_param)
+        parameters = bytearray(struct.pack(">BBB", self.COTP_PARAM_PDU_SIZE, 1, self.tpdu_size))
+        if self.calling_tsap is not None:
+            parameters.extend(struct.pack(">BB", self.COTP_PARAM_CALLING_TSAP, len(self.calling_tsap)))
+            parameters.extend(self.calling_tsap)
+        if self.called_tsap is not None:
+            parameters.extend(struct.pack(">BB", self.COTP_PARAM_CALLED_TSAP, len(self.called_tsap)))
+            parameters.extend(self.called_tsap)
+
+        pdu_length = 6 + len(parameters)
         base_pdu = struct.pack(
             ">BBHHB",
             pdu_length,  # PDU length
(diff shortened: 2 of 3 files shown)


gijzelaerr:
@esternu Could you rerun the two pure-Python compatibility reproducers from #889 and #890 against commit c9988603cac0eb93501f89cb9c65dee15671df24? This PR changes the COTP connection-confirmation framing and fragmented/coalesced request handling specifically around those failures. No physical PLC is needed; a pass or remaining traceback from both reproducers would close the validation loop.

esternu:
Thanks for the quick fix! I reran both against c998860 - **both pass**.

**Setup:** pure-Python server from the PR checkout, no PLC. Two clients: Telegraf 1.30.3 `inputs.s7comm` (gos7), i.e. the client from the original reports, and a small raw-socket script with deterministic timing (below). `tests/test_server.py`: 77 passed.

| | before (4984721) | c998860 |
| --- | --- | --- |
| #889 Connection Confirm length | 14 bytes | **22 bytes** |
| #889 Telegraf connects and reads | `ISO : Invalid PDU received` | reads values |
| #890 header just before / rest just after the deadline | `Invalid TPKT version: 2`, connection reset | valid response |
| #890 Telegraf polling every 5 s for 65 s | 3 aborts, `Invalid TPKT version: 2` (measured with only the #889 part applied, since plain master cannot connect) | 13 reads, 0 errors |

**One edge case you may want to look at - not a blocker for #889/#890.** A request split into two COTP fragments, with a gap longer than `RECEIVE_DEADLINE` before the second fragment's TPKT header arrives. With c998860 the header wait for the second fragment starts a fresh deadline and then raises a plain `TimeoutError`. The client handler treats that as idle (`except socket.timeout: continue`), so the already received first fragment is silently dropped, the connection stays open and the client never gets a response. That is the same behaviour as before the PR. Since the header of a *subsequent* fragment is by definition mid-request, raising `S7ConnectionError` there - as the PR already does for the payload read - would close the connection explicitly instead. Real clients rarely stall that long between fragments, so this is mostly about consistency with the "prevent stream desynchronisation" goal. A minimal variant:

```python
        started = False
        while True:
            header_deadline = time.monotonic() + self.RECEIVE_DEADLINE
            try:
                tpkt_header = self._recv_exact(4, header_deadline)
            except TimeoutError as e:
                if started:
                    raise S7ConnectionError("Receive deadline exceeded between COTP fragments") from e
                raise
            started = True
```

**About your comments on #889 and #890 asking for PRs:** since #893 already covers both, I don't want to open duplicates. I do have fixes for both on my fork - `esternu:fix/#889-cotp-cc-tsap-echo` and `esternu:fix/#890-receive-deadline-idle` - and the #890 one covers the fragment case above, with regression tests for all three timeout situations (request arriving as the idle wait runs out, timeout after the header, idle timeout without any byte). Happy to open them, or to contribute just the tests to #893, whichever you prefer.

gijzelaerr:
Thanks for the detailed validation and reproducer. I added the between-fragments timeout handling and a regression test in 363374c, and merged current master into this branch. The PR is conflict-free and all checks are green.

